PR #152, opened 2018-03-15
Added description of mode parameter in NoddingBackgroundModule

gcugno:
- Added description of the mode parameter in the NoddingBackgroundModule


tomasstolker:
Thanks! Made a small improvement. Default values are given in the function \_\_init\_\_ so have not to be provided in the parameter description.

PR #154, opened 2018-03-15
extra_rot in clockwise direction

gcugno:
- extra_rot acts in the clockwise direction

- Proposal: I would name the RotateAndStackModule DerotateAndStackModule. Rotation to me means that every frame is simply rotated by the same angle, while derotation means (to me) that at the end all frames have the same orientation.. @tomasstolker What do you think?


tomasstolker:
Yes I agree, changed the name to DerotateAndStackModule.

PR #156, opened 2018-03-16
Made a few changes in the module and parameter descriptions

gcugno:
Just made a few changes so that all the module descriptions are similar. In the RemoveLinesModule part of the description came from the old CutTopLinesModule


tomasstolker:
👍

PR #161, opened 2018-03-19
Added SDIPreparationModule

gcugno:
- Module that prepares the continuum images for the subtraction from the line filter images. 

- Still to be tested on Halpha dataset

- the images are normalized by the filter width, then upscaled and finally cropped to the initial dimensions

- This module has to be combined to the SubtractionImagesModule. Between them an alignment between the two stacks is necessary

- I don't like the fact that there is a new file with only this module. Maybe we can add it to the PSF Subtraction.py? We can discuss it later..

### DIFF
--- a/PynPoint/ProcessingModules/FrameSelection.py
+++ b/PynPoint/ProcessingModules/FrameSelection.py
@@ -8,6 +8,7 @@ import warnings
 import numpy as np
 
 from astropy.nddata import Cutout2D
+from photutils import geometry
 
 from PynPoint.Core.Processing import ProcessingModule
 from PynPoint.ProcessingModules.StarAlignment import StarExtractionModule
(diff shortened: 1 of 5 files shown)


tomasstolker:
Something seems wrong with the commit? Only __init__.py is changed.

gcugno:
- SDIPreparationModule tested. It is now in the PSFpreparation.py file

- RescaleImagesModule is now able to scale up the flux and only the image dimension (was needed for the SDIPreparationModule)

- 

- In FrameSelection.py there is the StellarPhotometryModule, which allows to look for variations in the counts in each frame, by simply counting them and writing them for each image within a given area. This should allow a simple statistical overview of the data
- updated __init__.py

Any comment is welcome!

tomasstolker:
Thanks a lot @gcugno.

There seems to be something wrong when I look at "Files changed". Some the code seems merged in the middle of SortParangModule? And there are still residuals from merge conflicts that you tried to solve? The CI also fails with the build. Please try to fix that or if needed start on a new branch and create a new pull request. Let me know if I can help.

In StellarPhotometryModule we could maybe also use the (circular) aperture photometry functionalities of photutils?

PR #163, opened 2018-03-20
SDIPreparationModule and update RescaleImagesModule

gcugno:
- Prepared, and tested the SDIPreparationModule

- Updated RescaleImagesModule. Now it is possible to multiply the image flux by a given factor.

- Updated __init__.py


tomasstolker:
Great, thanks. The CI seems to be failing, could you check what is wrong?

PR #164, opened 2018-03-20
StretchImagesModule

gcugno:
Module to stretch the images given by a given factor.


tomasstolker:
Is this the same zoom function as in PSFpreparationModule (ie scipy.ndimage.interpolation.zoom)? Because that is for resampling and not for stretching, right?

PR #165, opened 2018-03-21
StellarPhotometryModule

gcugno:
module that outputs the number of counts per image within a centered circular region of given radius.


gcugno:
Done!

PR #166, opened 2018-03-23
MeanStackModule

gcugno:
- Small bug I forgot to correct in AperturePhotometryModule

- MeanStackModule, which outputs the mean of the images of the input stack. It is necessary for the creation of a single photometry frame for AGPM data


gcugno:
uhmmm.. You're right... 
I was sure that there was something but could not find it... I tried with StackandSampling and MeanCube but did not work ;-)

Not related question: is there a reason why AddLinesModule only supports 3dim images? At the end of the preparation of the photometry master frame, I have to make PHOTOMETRY and SCIENCE images of the same dimensions, and therefore empty lines should be added.

tomasstolker:
> Not related question: is there a reason why AddLinesModule only supports 3dim images? At the end of the preparation of the photometry master frame, I have to make PHOTOMETRY and SCIENCE images of the same dimensions, and therefore empty lines should be added.

No there is not. All processing modules should support both 2D and 3D images but probably not all of them will work (yet) when the input tag is 2D.

tomasstolker:
DerotateAndStackModule uses now the MEMORY attribute such that only stacking and no derotation is faster.

PR #185, opened 2018-04-11
Small fixes in CropImagesModule, AngleCalculationModule, DerotateAndS…

18alex96:
Small fixes in following modules:
- CropImagesModule: x and y must be switched, when cropping around the center of an image (_center=None_). Tested on non-quadratic SPHERE/IRDIS images (pixelscale: 2048x1024)
- AngleCalculationModule: pupil_pos must be an array filled with zeros in case of SPHERE data. Otherwise an error is raised, when the pupil position is calculated.
- DerotateAndStackModule: In case of _derotate=True_ the variable _im_rot_ is overwritten each step of the for loop without using it. This way it is summed up, so that the average can be calculated correctly afterwards.
- StarAlignment: The old way of importing the reference images creates a 4 dimensional array, which results in problems later on.

### DIFF
--- a/PynPoint/ProcessingModules/ImageResizing.py
+++ b/PynPoint/ProcessingModules/ImageResizing.py
@@ -74,7 +74,7 @@ class CropImagesModule(ProcessingModule):
                 if size > image_in.shape[0] or size > image_in.shape[1]:
                     raise ValueError("Input frame resolution smaller than target image resolution.")
 
-                image_out = image_in[y_off:y_off+size, x_off:x_off+size]
+                image_out = image_in[x_off:x_off+size, y_off:y_off+size]
 
             else:
                 x_in = int(center[0] - size/2)
--- a/PynPoint/ProcessingModules/PSFpreparation.py
+++ b/PynPoint/ProcessingModules/PSFpreparation.py
@@ -519,10 +519,10 @@ class AngleCalculationModule(ProcessingModule):
             pupil_pos = self.m_data_in_port.get_attribute("PUPIL")
 
         elif self.m_instrument == "SPHERE/IRDIS":
-            pupil_pos = None
+            pupil_pos = np.zeros(np.size(steps))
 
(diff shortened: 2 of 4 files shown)


tomasstolker:
Nice work, thanks! Made a few small changes, let me know if you agree.

> DerotateAndStackModule: In case of derotate=True the variable im_rot is overwritten each step of the for loop without using it. This way it is summed up, so that the average can be calculated correctly afterwards.

Rewrote this module a bit, should work properly now for all combinations of derotate=True/False and stack=True/False

> StarAlignment: The old way of importing the reference images creates a 4 dimensional array, which results in problems later on.

Could you elaborate a bit? The ndim of the references images is checked with a separate case for ndim=2 and ndim=3. How can it be that ndim of ref_images is 4?

18alex96:
> DerotateAndStackModule: In case of derotate=True the variable im_rot is overwritten each step of the for loop without using it. This way it is summed up, so that the average can be calculated correctly afterwards.

Looks good!

> Could you elaborate a bit? The ndim of the references images is checked with a separate case for ndim=2 and ndim=3.

Yes, that is corrrect!

> How can it be that ndim of ref_images is 4?

`if ndim == 3` and the following if-statement is `False` it comes to the `else` expression, where the reference image is currently loaded as `np.array(self.m_ref_image_in_port.get_all())`. Since the dimension of `self.m_ref_image_in_port.get_all()` is already 3, the additional cast to an `numpy` array creates a 4 dimensional object. Or did I do something wrong?
I realized this behavior, when checking later for `ref_images.ndim`. None of the two cases `ref_images.ndim==2 or ref_images.ndim==3` applied, since `ref_images.ndim==4`.

tomasstolker:
Ok yes I see what you mean now, seems correct!

Additionally I removed the ndim condition when register_translation is called which is not needed I think because ref_images is always 3D, also if the original reference image is 2D.

PR #208, opened 2018-04-19
fix FluxAndPos

gcugno:
- PSF size was not looking for the PSF size, but the input frames size


tomasstolker:
Thanks, well spotted :).

Did you test the minimization module with the normalization set to True in the PSF preparation? Personally I would not use it with photometric measurements.

gcugno:
Mhm sorry, I tested both of them and didn't notice that it remained set to True. I agree that norm=False is the right way to go for photometric studies..

PR #249, opened 2018-07-23
Small fix: Instead of star_data the reference data needs to be reshaped.

18alex96:
Minor fix: `ref_star_data` needs to be reshaped instead of `star_data`.


tomasstolker:
Thanks, @18alex96! That should indeed be `ref_star_data`. I guess that means that you are the first one doing RDI instead of using the star data as reference...

18alex96:
Yes, I was using PynPoint for RDI for quite some time already. Just recently switched to the new Pca module. I will let you know, when I encounter further issues!

PR #251, opened 2018-08-08
dither parameter in WaffleCenteringModule

tomasstolker:
- Added a _dither_ parameter (default is False) in the WaffleCenteringModule. If set to True then the DITHER_X and DITHER_Y attributes will be used to apply an additional correction for a dithering offset. In the config file DITHER_X should be set to ESO INS1 DITH POSX and DITHER_Y to ESO INS1 DITH POSY for SPHERE/IRDIS data. Tested on an IRDIS pupil-stabilized DPI data set.

- Updated the requirements.txt file

### DIFF
--- a/PynPoint/ProcessingModules/StarAlignment.py
+++ b/PynPoint/ProcessingModules/StarAlignment.py
@@ -720,26 +720,31 @@ class WaffleCenteringModule(ProcessingModule):
         :return: None
         """
 
+        def _get_center(ndim, center):
+            if ndim == 2:
+                center_frame = self.m_center_in_port.get_all()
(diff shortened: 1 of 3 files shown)


tomasstolker:
The module also requires the NFRAMES attribute (typically set to NAXIS3 in the config file), in order to select the DITHER_X and DITHER_Y values belonging to the images from a FITS cube.

tomasstolker:
- The _center_ parameter can be set to None in the WaffleCenteringModule in which case the center of the image is used.
- The Util.ModuleTools.image_center supports both square and rectangular images.

PR #259, opened 2018-08-23
Introduce parameter for mean subtraction before PCA

18alex96:
Introducing new parameter `subtract_mean` to control, whether the average of each stack is subtracted before the principal components are calculated.

### DIFF
--- a/PynPoint/ProcessingModules/PSFSubtractionPCA.py
+++ b/PynPoint/ProcessingModules/PSFSubtractionPCA.py
@@ -272,26 +272,25 @@ class PcaPsfSubtractionModule(ProcessingModule):
         # get all data
         star_data = self.m_star_in_port.get_all()
 
-        # subtract mean, if required
-        if self.m_subtract_mean:
-            mean_star = np.mean(star_data, axis=0)
-            star_data -= mean_star
-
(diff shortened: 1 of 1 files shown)


tomasstolker:
Thanks @18alex96!

Just to update the others about our offline conversation. The goal here is to make the PCA module also working for RDI in which case the tag of the reference images is different from the science images for example applicable if there is no or little field rotation.

The problem with the current implementation is that the mean image is subtracted from both the science and reference images. The first is done manually in the pipeline module but the latter is done within the PCA.fit function of scikit-learn which can therefore not be turned off. So this is a bit of a disadvantage of using the sciki-learn functionalities for the PCA analysis.

The new _subtract_mean_ parameter controls the subtraction of the mean from the science data but also computes the mean of the reference images, orthogonalizes it with respect to the PCA basis, and adds it as the first PC to the basis while the last PC is removed to keep the number of PCs fixed.

tomasstolker:
The CI fails to finish, probably due to 'UnboundLocalError: local variable 'mean_ref_star' referenced before assignment'. Could you check what is wrong @18alex96? The test case for the PSF subtraction should not cause any errors since the default value of subtract_mean is True.

PR #268, opened 2018-10-09
Noise-weighted residuals, Core.Attributes, AttributeReadingModule, and more

tomasstolker:
- Updated Makefile.

- Created Core.Attributes which contains a dictionary with all implemented attributes and some relevant information: section in the config file, data type, default value, and static/non-static. Pypeline is updated for the use of Core.Attributes.

- Simple set_attribute function in Pypeline which can set and static or non-static attribute of a database tag with the provided values. Existing attribute values are directly overwritten without further check.

- AttributeReadingModule to read a list of values from a text file and attach them as non-static attribute to a database tag. Only attributes from Core.Attributes can be written.

- FalsePositiveModule can take both a single image or a stack of images. So the SNR/FPF can be directly calculated on all residuals for the different numbers of PCs from the PCA module.

- Noise-weighted residuals (see Bottom et al. 2017) from the PSF subtraction are written to res_weighted_tag. The default is set to None, as the other output tags. Implemented for both single- and multiprocessing.

- Test cases for Hdf5WritingModule, Hdf5ReadingModule, and AngleCalculationModule


tomasstolker:
Progress dots added for the PSF subtraction with multiprocessing.

PR #275, opened 2018-10-17
For modules using an input PSF, individual PSF frames are averaged if PSF is 3D and doesn't match number of image frames

aboehle:
In the ContrastCurveModule, the warning about taking the average of the psf_in_tag (if the PSF is 3d size of this tag doesn't match the size of the image_in_tag) was being thrown, but the PSF averaging was not performed.  I added the PSF averaging to this module and also to other modules that use an input PSF (SimplexMinimizationModule and MCMCsamplingModule).


tomasstolker:
Thanks, I had forgotten to update the warning... Initially it was taking the average but I thought that it would be better if this is done separately from the limits and minimization modules in order to make the user aware of what the PSF template looks like. So I would like to suggest to update the warning to point the user to the DerotateAndStackModule which can be used to mean or median combine a stack of images. The advantage of this module is that the MEMORY keyword is used for calculating the mean. What do you think?

aboehle:
I see what you mean, I agree it's good for users to be able see that intermediate step of the final PSF template.  I will change the warnings!  Since most users will have multiple images of the reference PSF and need to take the average, I will also add this information to the module descriptions.

aboehle:
The changes are in now.  I also changed the name/description of the old DerotateAndStackModule to DerotateAndOrStackModule to make it more clear that the module can be used to only stack if needed.

tomasstolker:
Great, thanks a lot! The test cases are no longer finishing, could you check what is wrong? You can use "make test" in the PynPoint folder. Please also run pylint on the changed files if you haven't done so yet (only for easy improvements).

aboehle:
The tests run now! The MCMC test was failing because it was still using a 3D PSF for the input and then the new ValueError was raised.  I edited the test so that the PSF is averaged using the DerotateAndOrStackModule before being used in the MCMC.

Through this, I also noticed that the MCMC module was a different fake_planet function (which was still able to take a 3D PSF as input and then average it) than the other modules, so I also merged the two fake_planet functions.

PR #293, opened 2018-11-23
Improved masking PcaPsfSubtractionModule

tomasstolker:
I have started to rewrite the PcaPsfSubtractionModule such that pixel values that are set to zero are excluded from the PCA basis. Before, masked pixels are set to zero (so they do not affect the PSF subtraction) but they are still being processed which causes an unnecessary overhead. So far I only implemented this for single processing, but will do this for multiprocessing of the PcaPsfSubtractionModule, and the PCA function that is used by the ContrastCurveModule, SimplexMinimizationModule, and MCMCanalysisModule asap.

### DIFF
--- a/PynPoint/ProcessingModules/DetectionLimits.py
+++ b/PynPoint/ProcessingModules/DetectionLimits.py
@@ -247,10 +247,9 @@ class ContrastCurveModule(ProcessingModule):
                     im_shape = (fake.shape[-2], fake.shape[-1])
                     mask = create_mask(im_shape, [self.m_cent_size, self.m_edge_size])
 
-                    im_res = pca_psf_subtraction(fake*mask,
-                                                 parang,
-                                                 self.m_pca_number,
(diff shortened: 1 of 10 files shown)


tomasstolker:
- Improved masking for any pipeline module doing PSF subtraction. Pixels set to zero (in the science data) are excluded for the PCA basis and fit. The original images size is restored before derotation.

- Updated Util.PSFSubtractionTools.pca_psf_subtraction which can take a PCA object from scikit-learn as keyword argument.

- PcaPsfSubtractionModule uses Util.PSFSubtractionTools.pca_psf_subtraction both with single- and multiprocessing.

- Added Util.Residuals.combine_residuals which combines a stack of images with a "mean", "median", "weighted", or clipped. This functions is used by various pipeline modules after Util.PSFSubtractionTools.pca_psf_subtraction.

- Documentation for Util.MultiprocessingPCA.

PR #294, opened 2018-11-28
Python 2+3 compatibility

tomasstolker:
- Made all code compatible with Python 2 and Python 3.

- Updated .travis.yml and tox.ini.

- Updated versions in requirements.txt.

- Travis CI is testing Python 2.7, Python 3.6, and Python 3.7, both on osx and linux (all test cases OK).


tomasstolker:
Updated the use of coveralls which has to get the results from all tests on the CI, which are running in parallel for the different operating systems and Python versions.

The pull request is ready to be merged. Let me know if you have any comments or questions.

PR #300, opened 2018-12-20
Multiprocessing ContrastCurveModule, PEP 8 renaming package structure

tomasstolker:
- Renamed the folders and filenames in order to follow the PEP 8 naming conventions.

- The easiest way to import pipeline modules is, e.g. `from pynpoint import ContrastCurveModule`

- Note that `import PynPoint` has changed to `import pynpoint`.

- Updated documentation and examples.

- The `ContrastCurveModule` is now running the positions in parallel by setting `CPU` in the configuration file. I did not use the multiprocessing functionalities of PynPoint but the `multiprocessing` functionalities of Python directly. This was more straightforward and in this case possible because the database is not accessed during the calculation of the limits. All data is stored at once in the end.

- Tested with Python 2.7, 3.6, and 3.7 on OSX (!).

- Deprecated the `pca_out_tag` in `ContrastCurveModule`, which was storing all PSF subtraction residuals. In principle this can still be implemented with the multiprocessing but was easier to leave out.

Let me know if you have any comment or questions.


tomasstolker:
Multiprocessing for the PSF subtraction is working on OSX now. Tested with Python 2 and 3 with the dependency versions of requirements.txt. Updated the `PcaPsfSubtractionModule`.

PR #304, opened 2019-01-21
Pullrequest Sharedmemory

404NotFoundv2:
I think this is the way to create a pull request.
There might be a better way to implement shared memory but this seems to work to my knowledge.
I will run some additional testing.


tomasstolker:
Thanks for the pull request! Could you tick the "Allow edits from maintainers"?

tomasstolker:
Added sharedmem to the requirements.txt. All test cases are OK.

PR #307, opened 2019-02-10
ContrastCurveModule

404NotFoundv2:
Here is the implementation of multiprocessing for the ContrastCurveModule. I tested it both on Bluesky & Euler.

The input arrays containing the -_images_- and -_psf_- are saved in numpy binary format in the -_working_place_-. These numpy files are loaded in each subprocess. When all the processes have finished, those files are removed.

The number of processes is set by the -_cpu_- in the config file. As each process gets an copy of this in memory, the cpu should not be set too high.

Also the '.' has been replaced by a percentage sign, but this can easily be, if required, removed.

Lastly, the result in each process is saved in a queue. This object is guarantees safe handling of information between multiple threads/processes.

I left some comments in there to understand what each part does.

### DIFF
--- a/pynpoint/core/pypeline.py
+++ b/pynpoint/core/pypeline.py
@@ -197,6 +197,8 @@ class Pypeline(object):
                 if attributes[key]["value"] is not None:
                     config.attrs[key] = attributes[key]["value"]
 
+            config.attrs["WORKING_PLACE"] = self._m_working_place
+
             hdf.close()
 
         config_file = self._m_working_place+"/PynPoint_config.ini"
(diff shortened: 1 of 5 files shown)


tomasstolker:
Thanks for the efforts, @404NotFoundv2! It looks good to me.

The 'dot' was used as progress indicator because processes may not finish in order. Is the percentage indicator always increasing?

404NotFoundv2:
Great!
The dot was written to the output when a magnitude at a single position had been found/converged, correct?

This is how what I've done now:
Say _cpu=5_, number of positions to calculate=40
It starts the calculating the first 5 positions. When all 5 are finished, the percentage sign jumps from 0% to 12.5%. Thus, if cpu would be set to 40, it jumps from 0 to 100%.

The dot would output a '.' when a single process is finished, but usually the 5 processes finish at approximately the same time.

TL;DR, Yes it increases. When #*_cpu_* process is finished, it increases

tomasstolker:
OK good, thanks!

PR #312, opened 2019-02-20
Bug fix AngleCalculationModule

18alex96:
- fix bug concerning header keywords, when reading FITS files
- improve AngleCalculationModule for SPHERE data
- remove ephem as requirement

### DIFF
--- a/.idea/PynPoint.iml
+++ b/.idea/PynPoint.iml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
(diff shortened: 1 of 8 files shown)


tomasstolker:
Thanks a lot @18alex96!

- There are several files added within the hidden .idea folder (probably by PyCharm?). Could you remove those? You can add the .idea folder to .gitignore if needed.
- What was the reason that the version of six changed from 1.12 to 1.11 in the dependencies?
- Somehow the CI was not triggered by the PR, perhaps something went wrong on the side of Github or Travis. Can you close the current PR and open a new one?

18alex96:
- Yes, this is due to PyCharm and I will ignores these
- No idea how this has happened, I changed it back to version 1.12
- I just closed this pull request and will create a new one immediately.

PR #313, opened 2019-02-21
Angle calculation fix

18alex96:
- fix bug concerning header keywords, when reading FITS files
- improve AngleCalculationModule for SPHERE data
- remove ephem as requirement (sideral time calculation done by astropy)


tomasstolker:
Thanks! The CI is working now.

The test cases of `test_psfpreparation` cause an error, could you have a look and update them if needed? You can run them with `pytest test_psfpreparation.py` in `tests/test_processing`.

18alex96:
I checked the test case and there are two main reasons, why it fails (as far as I can see).
1. I slightly changed the calculation of the sideral time. Before, I was using `ephem`, but I changed this to `astropy` instead, which seems to be a more elegant implementation (and we do not need another package requirement). Furthermore, I suppose that `astropy` is kept more up to date compared to `ephem` (larger community is using it). This change, however, resulted in a slight adjustment of the computed sideral time -> leading to a slight change in parallactic angle -> causing the test case to fail: 
- ephem (old implementation): `data[0]=-55.04327383` degrees
- astropy (new implementation): `data[0]=-55.04109752` degrees
I suppose `ephem` is not taking into account the latest models for the earth's precession and nutation, which `astropy` claims to do. For that reason, I would trust the new implementation (the difference between both methods is below our astrometric precision, anyway).
2. Unfortunately, the correct RA/DEC parameters in the SPHERE header have another format (hhmmsss/+-dddmmss) than the RA/DEC parameters used before (degrees). That is why I had to implement a new way of converting this input to degrees (as it was general input before and as the NACO input still is). To clarify this for the user, I implemented a warning that is raised when calculating angles for any instrument of SPHERE. For this reason, we have to define different test cases for SPHERE and NACO data, respectively.
Summarizing my thoughts, I would suggest to:
- Adapt the NACO test case so that it is checking for the correct values that were calculate using `astropy`
- Re-define the SPHERE test cases, so that the input values for RA and DEC match the corresponding SPHERE header keywords' format (and of course do the same update taking into account the slight difference when computing the parallactic angles with astropy)
@tomasstolker Let me know what you think about this suggestion and I can try to implement it.

tomasstolker:
Ok yes sounds goods, thanks the improvements, please adjust the test cases as you suggested.

tomasstolker:
What Python version do you use? Testing by the CI with Python 3 failed while Python 2 was ok...

18alex96:
Indeed, I am using Python 2, but I do not really see, why the changes should not work for python 3 (although I am not an expert for that).
Looks like the error is raised in line 497 of `psfpreparation.py`, where we try to initialize an `astropy.tim.Time` object. Apparently, the format of the date is not recognized, which is`val = array(b'2012-12-01T07:09:00.0000', dtype='|S24')`. 
@tomasstolker Stupid question: should it not be rather `val = '2012-12-01T07:09:00.0000'` (without the array) or something like that? Are you aware of any python 3 feature that may cause this issue (as I am really no expert on python 3?

tomasstolker:
Python 2 and 3 handle byte strings and unicode strings differently. It is fixed now so I will merge this PR.

PR #315, opened 2019-02-27
optimize parameter FalsePositiveModule

tomasstolker:
- `optimize` parameter for `FalsePositiveModule`. If set to True, the FPF is minimized by leaving the position of the planet aperture as free parameter. The aperture radius is kept fixed. The absolute tolerance is set to infinity and the fractional tolerance on the FPF to 1e-6.
- `util.image.cartesian_to_polar` function to convert pixel coordinates to polar coordinates.
- `util.image.get_image` function to select an image from a database tag by its index number.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -23,7 +23,7 @@ The pipeline works with two different components:
 2. The actual pipeline :class:`pynpoint.core.pypeline` which capsules a list of pipeline modules.
 
 .. important::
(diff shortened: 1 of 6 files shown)


tomasstolker:
The fractional tolerance parameter of the `minimize` function of `scipy` has been deprecated. Therefore I have added an `tolerance` keyword parameter in `FalsePositiveModule` which sets the absolute tolerance on the pixel position of the aperture (default is 0.01 pix) while the absolute tolerance on the FPF is set to infinity.

PR #319, opened 2019-04-02
Angle shift and shape check

tomasstolker:
- Additional correction of 360 deg in `AngleCalculationModule` if the calculated parallactic angles are smaller than -180 deg or larger than 180 deg, just to have them in a more intuitive range (which was not always the case because of the additional rotation offsets that are applied). @18alex96 could you have a brief look?

- Check in `CombineTagsModule` if the images of all datasets have the same size. If not, an error is raised.


18alex96:
The change in the `AngleCalculationModule` makes sense and looks good to me.
Thanks for optimizing this @tomasstolker

PR #320, opened 2019-04-02
Psf_in_tag in ContrastCurveModule doesn't accept 2D image

404NotFoundv2:
Hi Tomas,

I get an error when I use the contrastcurvemodule where the psf_in_tag is a single, 2D image (512, 512). This is because the fitsreadingmodule converts the (512, 512) array into (1, 512, 512). I changed the fitsreadingmodule such that it keeps the 2D size if there is only a single fits file found in the folder and that fits file contains only 1 image.

Another way would be to  change the way the psf_in_tag  in the contrastcurvemodule reads the image and allow it to accept (1, 512, 512) and read it as a single image.

The error:
ValueError: The number of frames in psf_in_tag does not match with the number of frames in image_in_tag. The DerotateAndStackModule can be used to average the PSF frames (without derotating) before applying the ContrastCurveModule.

Jasper

### DIFF
--- a/pynpoint/processing/limits.py
+++ b/pynpoint/processing/limits.py
@@ -186,11 +186,11 @@ class ContrastCurveModule(ProcessingModule):
 
         self.m_aperture /= pixscale
 
-        if psf.ndim == 3 and psf.shape[0] != images.shape[0]:
-            raise ValueError('The number of frames in psf_in_tag does not match with the number of '
-                             'frames in image_in_tag. The DerotateAndStackModule can be used to '
-                             'average the PSF frames (without derotating) before applying the '
-                             'ContrastCurveModule.')
+        if psf.shape[0] != 1 and psf.shape[0] != images.shape[0]:
+            raise ValueError('The number of frames in psf_in_tag {0} does not match with the '
+                             'number of frames in image_in_tag {1}. The DerotateAndStackModule can '
+                             'be used to average the PSF frames (without derotating) before '
+                             'applying the ContrastCurveModule.'.format(psf.shape, images.shape))
 
         pos_r = np.arange(self.m_separation[0]/pixscale,
                           self.m_separation[1]/pixscale,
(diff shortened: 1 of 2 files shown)


tomasstolker:
Thanks for pointing that out, @404NotFoundv2.

I am not sure what is the best approach here and it would be good to think about this more generally. Storing single images as a 3D dataset (with the first dimension set to 1) may have the advantage that the first dimension can always be used to infer the number of images in the stack. This is probably the reason why @majobodo implemented it in this way in `FitsReadingModule`?

But, there are several modules, like the `ContrastCurveModule` which distinguish between 2D (i.e. 1 image) and 3D arrays (i.e. more than 1 image). This could be consistently adjusted to the first approach though, if we think that that would be better.

Personally, I am thinking that the first approach might actually be better, because the implementation would be a bit more straightforward, not always having to consider the 'special' case of a dataset with 1 image (i.e. 2D data) separately from the 3D case. So this would mean that single images are always stored as 3D dataset.

I would be interested to hear other opinions on this. @spquanz @aboehle @majobodo @18alex96 @gcugno

404NotFoundv2:
To keep the structure consistent (3-dim), I would agree to keep the fitsreadingmodule as it is and fix the _psf_in_tag_ error in the contrastcurvemodule so it accepts a single image (and the same number of frames as the _image_in_tag_). This will probably also avoid errors in the future, as the # of frames is always the first dimension.

Ps. I reverted my commit, but I didn't know the issue would automatically be closed. I will check if I can change the limits.py file and reopen this issue.

gcugno:
I agree with @404NotFoundv2, fixing the psf_in_tag error makes probably more sense. In this way, the first dimension shows always the same thing, i.e. the number of images.

tomasstolker:
Ok agreed, let's work only with 3D image arrays in PynPoint, both for ``nimages = 1`` and ``nimages > 1``. I will see what I can do to check the other modules.

PR #322, opened 2019-04-03
NEAR documentation

404NotFoundv2:
Hi Tomas,

I've created this pull request so we can both change and edit the near page. The build works, but gives 3 duplicate label warnings, which I haven't fixed yet.

As you see, I've put the near installation file below the _User Documentation_ header, but feel free to change it.


tomasstolker:
Thanks for the PR @404NotFoundv2. I think that that is a good location in the docs. I have subdivided the text in two pages. We could also add some general information about NEAR and the data format. Let's discuss.

404NotFoundv2:
Would it be useful to add a sentence in the installation instructions that it is recommended to use python3? As the commands described there will use python2 if python3 is not installed on the system.

tomasstolker:
I have added a sentence in the installation section of the docs that Python 3 is recommended. I have also updated the NEAR documentation.

PR #324, opened 2019-04-05
Converting 2D to 3D arrays

tomasstolker:
Following #320, I have consistently updated the pipeline modules such that single images are always stored in a 3D array with the first dimension set to unity.


tomasstolker:
Added the ``StackCubeModule`` which will replace the ``MeanCubeModule`` in a future release.

PR #327, opened 2019-04-11
New ContrastCurveModule

tomasstolker:
I implemented a new `ContrastCurveModule` which does not require an iteration at each position. Each position is still processed separately (and in parallel) but the self-subtraction is calculated once for which a fake planet is injected based on the local noise level. This makes the computation less complex and more stable, and also does not require the `accuracy` and `magnitude` parameters so it is more straightforward to run.

I ran a test on a NACO NB3.74 dataset of beta Pic which shows very similar results. With the old module, I used an accuracy of 0.1 (so the iteration stops if the FPF is within 10% of the threshold, after which an interpolation is applied) which typically means 5-10 iterations per position. Therefore with the new module this will run 5-10 times faster.

limits


tomasstolker:
And here a comparison of the old and new `ContrastCurveModule` implementation with the NEAR test data:

limits

PR #328, opened 2019-04-11
Multiprocessing Queue Order Fix

benassa-de-glassa:
Due to multiprocessing the results from the queue may not be in the same order as the processes were created and started. Therefore the current way of summarizing the results for the out_port may lead to incorrect final results (especially when many cores and few different angles are used). 
I mitigated this by storing the data in a dictionary using the separation as keys.

Here the output as given by the queue (with the separation in the first, the angle in the second, the magnitude in the third and the fpf in the fourth column):
image

Here the incorrect ContrastCurve result (the contrast magnitude is 6.34 and 6.15 instead of 4.98 and 7.51):
image

Here my result: 
image


tomasstolker:
Thanks a lot for looking into this @benassa-de-glassa. I do remember testing the order of the results when we started implementing the multiprocessing for the `ContrastCurveModule` so I am a bit surprised that this bug has appeared... Is this something you looked into @404NotFoundv2?

The fix seems good but I was wondering if an easier solution would not be simply ordering the results in the end by separation? Let me know what you think.

tomasstolker:
According to the Python documentation the `Queue` class works in a FIFO (first in first out) manner, so the first tasks added are the first retrieved. That would mean that the order should be conserved? I am probably misunderstanding the documentation...

https://docs.python.org/3/library/queue.html

Which Python version are you using @benassa-de-glassa?

benassa-de-glassa:
> The fix seems good but I was wondering if an easier solution would not be simply ordering the results in the end by separation? Let me know what you think.

Yes, that might work as well. I used the first implementation that came to mind. 
My implementation has the advantage that it does not depend on the number of angles per separation. So (in future) you could run many more angles in a separation range that you are particularly interested in. Also one could set a timelimit to the processes. 

> Which Python version are you using @benassa-de-glassa?
I'm using the python version 3.6.7. 
I think the Queue works indeed in a FIFO manner, but if a 'contrast_limit' process finishes earlier than an other it will also write to the Queue earlier. Thereby causing the Queue not to be ordered as expected.

404NotFoundv2:
Hmm this is indeed a bit surprising. I looked into this and this did not occur in my case.

The Queue object is indeed a FIFO object, but the problem here is that processes put their results in the queue which are not sorted by radius & angle (they usually are but can finish earlier).

I agree that the 'easiest' solution would be to sort the results on radius & angle, but I'm not sure what would happen if the contrast doesn't converge at a point. Does this give a NaN or does it not give anything?

benassa-de-glassa:
@404NotFoundv2 yes, if the contrast_limit function does not converge within 50 iterations, it does put np.nan as the magnitude in the queue

tomasstolker:
Thanks for the clarifications @benassa-de-glassa and @404NotFoundv2. I understand now what the problem is...

I made a few changes to the sorting of the results. @benassa-de-glassa, could you please double check that it is correct?

@404NotFoundv2: The return statement in `util.limits.contrast_limit` is no longer needed with the multiprocessing, right? I have commented it out.

404NotFoundv2:
> The return statement in util.limits.contrast_limit is no longer needed with the multiprocessing, right? I have commented it out.

Correct, that's not needed anymore.

benassa-de-glassa:
> I made a few changes to the sorting of the results. @benassa-de-glassa, could you please double check that it is correct?

@tomasstolker it looks correct to me and works in my case. 

I'd still like to argue that the dictionary method offers more flexibility, as it does not require the same number of results for all separations. 
This is useful when there are single processes which stall (this happens to me sometimes). Then the single stuck process can be terminated without making the entire Module run "worthless". I did this by setting a 8 hour timelimit. But this is a separate issue.

In my case this looks like:
image

tomasstolker:
Thanks for testing! I appreciate the efforts you made to make the module work for a changing number of position angles per separation, but I think that at this point it would be best to not add further complexity to the `ContrastCurveModule`. In particular because I plan to replace it with a more robust implementation which does not require the iteration (see #327).

Do you know why the processes stall sometimes? This never happened to me...

benassa-de-glassa:
> Do you know why the processes stall sometimes? This never happened to me...

I think its a particular issue with a certain dataset. I haven't figured out where exactly the problem lies. I'll keep you posted.

PR #330, opened 2019-04-17
Additional fits io options from txt file

benassa-de-glassa:
@tomasstolker this is my current implementation to add filenames from a txt file to the list of files to be read into a certain tag. 

I find it particularly useful when running dfits/fitsort to create a filelist which displays certain fits headers
`dfits "folder_with_fits_files"/*.fits | fitsort "list of interesting header keywords" > files.txt`
I then select the files I am interested in by hand and remove the header keywords and the fitsort header

I will add the possibility to pass a list next.


tomasstolker:
Great, thanks @benassa-de-glassa, this is a very useful feature for the `FitsReadingModule`. To limit the number of input parameters, I want to propose to merge the two parameters in a single `filenames` parameter. Then we can have three options:

- If the value is set to None (default) then the`input_dir` is used to read in FITS files.
- If the value is a string, then a text file is read, containing FITS files relative to the working folder.
- If the value is a list, then the filenames are read directly from this list (also relative to the working folder).

In the latter two cases the `input_dir` should probably be ignored.

benassa-de-glassa:
@tomasstolker I have implemented the single filenames parameter as you described above.. it works for me, but please double check

> Then we can have three options:
> 
>     * If the value is set to None (default) then the`input_dir` is used to read in FITS files.
> 
>     * If the value is a string, then a text file is read, containing FITS files relative to the working folder.
> 
>     * If the value is a list, then the filenames are read directly from this list (also relative to the working folder).
> 
> 
> In the latter two cases the `input_dir` should probably be ignored.

I haven't tested any weird cases like strings that contain " or ', but I think python will handle those on its own.

tomasstolker:
I have added a test case for the `filenames` parameter and made a few small changes. Please have a brief look @benassa-de-glassa before we merge it. Thanks!

pynpoint/readwrite/fitsreading.py
```diff
@@ -290,36 +288,36 @@ def _extra_attributes(self,
...
         if isinstance(self.m_filenames, str):
             files = self._txt_file_list()
-            location = ''
-        elif isinstance(self.m_filenames, list):
+            location = self._m_config_port.get_attribute("WORKING_PLACE")
+
+        elif isinstance(self.m_filenames, (list, tuple)):
             files = self.m_filenames
-            location = ''
+            location = self._m_config_port.get_attribute("WORKING_PLACE")
```
benassa-de-glassa: This way the `location` will become the working directory of the pipeline. 
I don't think that is a good idea, as the user should ideally not interact with the working directory at all.

benassa-de-glassa:
The changes look good for me, except for the location change to the `WORKING_PLACE` of the Pipeline. I personally would not let the user interact with that directory.

tomasstolker:
To summarize our earlier discussion... These are some possibilities:

- The FITS files are specified relative to the PynPoint default input folder or PynPoint working folder.

- The FITS files are specified relative to the Python working folder. Perhaps this may cause a problem if the script is executed from a notebook over ssh? I have no experience with that...

- The absolute paths are provided. This has a disadvantage in combination with dfits, which only writes relative paths. A workaround would be providing the full path of the text file (which contains the list of FITS files) and having the text file in the same folder as the actual FITS files as requirement.

tomasstolker:
Or another option would be using the path where the script with the pypeline code of the user is located, i.e. `os.path.dirname(os.path.abspath(__file__))`. Perhaps that would be a good solution?

benassa-de-glassa:
>     * The FITS files are specified relative to the PynPoint default input folder or PynPoint working folder. 
* As I have said earlier, I dislike the idea of using the PynPoint working folder on a conceptually.
>     * The FITS files are specified relative to the Python working folder. Perhaps this may cause a problem if the script is executed from a notebook over ssh? I have no experience with that...
* by Python working folder, do you mean the $PYTHONPATH, if so, I think that is a bad idea, who even knows their $PYTHONPATH? I have no idea how notebooks work via ssh either.
>     * The absolute paths are provided. This has a disadvantage in combination with dfits, which only writes relative paths. A workaround would be providing the full path of the text file (which contains the list of FITS files) and having the text file in the same folder as the actual FITS files as requirement.
* absolute paths are the simplest to implement, but might be cumbersome for users

> Or another option would be using the path where the script with the pypeline code of the user is located, i.e. `os.path.dirname(os.path.abspath(__file__))`. Perhaps that would be a good solution?

I think that this is the simplest and most intuitive solution.

tomasstolker:
The Python working folder would be the path where Python is executed (which can be different from the folder where the Python code is located).

Let's go for the location of the user's pypeline script then. I will make the changes.

PR #333, opened 2019-04-23
Restructure PSFpreparationModule

timothygebhard:
This pull request introduces several major changes to the `PSFpreparationModule`:

* The `resize` keyword has been deprecated, since resizing should take place during the alignment / centering stage; not during the PSF preparation. In order to not break the API, the keyword was not yet removed; however, its value will be ignored and a `DeprecationWarning` is raised if a non-default value is passed.
* Instead of using a `for`-loop to process as images in the stack separately, all operations are now using numpy's vectorization feature. On larger data sets, this leads to a *significant* performance improvement. 
**Example:** When processing a data set with 29681 frames, the time to run the `PSFpreparationModule` went down from 985 seconds to just 35 seconds (speed-up factor of ~28).
* The `memory_frames()` function is used to split up the input frame into smaller chunks to obey the memory constraints of the pipeline.
* Default value for `norm` argument was changed from `True` to `False`, that is, normalization is turned off by default. (**Note:** I unfortunately forgot to include this in the commit message…)
* Image masking now happens *before* the normalization step.

These changes (e.g., deprecation of the resizing feature) of course also required some modifications of the corresponding unit tests.


tomasstolker:
Nice work @timothygebhard! Thanks!

I made a few small changes:

- Changed the line length to 100. We can discuss to shorten lines to 79, but then we should consistently change it throughout the code...

- All images that are stored in the database should have 3 dimensions, also with a single image. I have added a warning in case `ndim = 2` in which case a dimension is added. It seemed that `np.atleast_3d` added a dimension to the last axis, whereas the dimension should be added to the first axis. So I used `np.newaxis` instead.

- If there is no data stored at a database tag then appending data with `data_dim=3` should work, regardless of the number of iterations in the for loop and number of images that have to be stored.

Please have a look and let me know if you have any comments or questions!

A tip for your next pull request  :), if you create your changes on a separate branch then I don't have to push to your master ;).

PR #336, opened 2019-04-25
Mass limits Module

benassa-de-glassa:
Hey
so I spent today writing this module which converts contrast limits into mass limits using: 

- a COND model file (see attachment)
- apparent magnitude and distance
- age of the system
- filter in which the data was observed

I figured it could be helpful to others as well.
Cheers,

[model.AMES-Cond-2000.M-0.0.NaCo.Vega.txt](https://github.com/PynPoint/PynPoint/files/3117922/model.AMES-Cond-2000.M-0.0.NaCo.Vega.txt)


benassa-de-glassa:
also I noticed I started this branch from origin instead of upstream, I didn't mean to undo the changes to FitsReading, and I have forgotten to squash the commits.. sorry about that

18alex96:
Thanks for the implementation @benassa-de-glassa. So far, always did this offline (without PynPoint), but I think it is a good idea to include this into the existing framework.
Some input from my side that might be helpful:
- `host_star_magnitude` sounds a little misleading, as this parameter seems to take both the magnitude and the parallax. What about renaming this to `host_star_properties` or so?
- Personally, I would prefer to enter the `age` in Myr, as we are usually dealing with very young systems (but that is my personal preference, so no worries, if you disagree).
- I did not entirely follow the recent discussion on relative and absolute paths, but I would prefer to open the model from an absolute path (as I do not store my codes and data in the same (sub-)directories). Not sure, if this is supported by your current implementation and in case you have discussed this issue already in depth, feel free to ignore my comment.
- It looks like the method only supports ages that are listed in the COND model. The method I have written myself for the conversion to mass limits supports any age by using the closest ages available and interpolating linearly between both. Let me know, if you are interested in this functionality and I can assist you with the implementation (if needed).
- Last but not least: In principle the methods also works for other model families than COND (e.g. Dusty, BT-Settl). So I would rather use the general term 
> atmosphere model
 (or something similar) instead of constraining the apparent use case to COND models.

benassa-de-glassa:
@18alex96 Yes I too used to do it offline, but I figured it could be useful for others too. 

- You're right `host_star_magnitude` is misleading, it is a remnant of an earlier implementation. I will rename that.
-  The age in Gyr stems from the COND model giving the age in Gyr, I think I will implement a version that supports both.
- It does currently just support the ages from the COND model, because there is so much uncertainty in the age of the system I'm currently looking at that it didn't really matter. I'd be very glad if you could share your implementation.
- It currently only works for COND models because that is the one I'm using. Actually the biggest pain in the implementation was to get the data from the file in a somewhat "structured" format. If you have a table of the other models at hand I'd be thankful if you could share them with me and I'll try to implement them too and rewrite the docstring

tomasstolker:
Thanks for the PR @benassa-de-glassa. I like the idea of including this module.

- I don't think that we have to include the isochrone data in the repository, instead I agree with the suggestion by @18alex96 that we could make the module more generic, so not limiting it to a specific atmosphere model, metallicity, and instrument. It seems that all the PHOENIX-based isochrones are provided in the same format (https://phoenix.ens-lyon.fr/Grids/) so we could let the user download the isochrones from that webpage and have a `filename` parameter pointing to the isochrones file.

- There is a bunch of files attached to this PR which are probably from your Python installation or something. Please remove these.

- I also think that ages are better given in Myr.

- I don't think that we really concluded on the absolute vs. relative paths discussion. For the `FitsReadingModule` the `filenames` is currently implemented as relative path. Perhaps we could build in a check: if the string starts with "/" then an absolute path is used, if not then a relative path?

- I agree with the suggestion to interpolate both the age and magnitude.

benassa-de-glassa:
@tomasstolker 


>     * I don't think that we have to include the isochrone data in the repository, instead I agree with the suggestion by @18alex96 that we could make the module more generic, so not limiting it to a specific atmosphere model, metallicity, and instrument. It seems that all the PHOENIX-based isochrones are provided in the same format (https://phoenix.ens-lyon.fr/Grids/) so we could let the user download the isochrones from that webpage and have a `filename` parameter pointing to the isochrones file.

- The isochrone data is not intended to be included in the repository. It is supposed to be downloaded individually by the user and be pointed at by `model_file`. I just included that file in my originial PR to ensure we were talking about the same thing.
I did not know they were all in the same format. That makes it a lot easier to implement. Or rather there is no need for any change then, except the docstring.


>     * There is a bunch of files attached to this PR which are probably from your Python installation or something. Please remove these.

- I must have messed up the .gitignore. Sorry about that. 

>     * I also think that ages are better given in Myr.
- I have reworked the host star properties, now that information 

>     * I don't think that we really concluded on the absolute vs. relative paths discussion. For the `FitsReadingModule` the `filenames` is currently implemented as relative path. Perhaps we could build in a check: if the string starts with "/" then an absolute path is used, if not then a relative path?
- Should that not happen automatically? Except for when one is not executing the python file in one's current working directory..
>     * I agree with the suggestion to interpolate both the age and magnitude.
- I will look into this on the weekend

benassa-de-glassa:
I messed up badly... I tried deleting the files, and by mistake also deleted fitsreading.py.. I am not that good with git yet and can't figure why `git revert 9f153d6` does not work..

tomasstolker:
No worries, you can get the `fitsreading.py` file from the `upstream/master` (i.e. simply download it from the Github page) and copy it back in your branch.

benassa-de-glassa:
@18alex96  @tomasstolker I implemented your suggestions. 
The model data is now linearly interpolated between the two nearest time points given by the model.
I have run a test for my data and the result looks sensible. I'd appreciate anyone double checking my code.

tomasstolker:
Thanks a lot for updating the module!

I have a few additional suggestions:

- I would remove the`mag_app` flag and `age_unit` from the star properties dictionary. Providing an apparent magnitude and an age in Myr seems enough.

- Having two interpolation functions (`_interpolate_data_by_age` and `_interpolate_model`) seems a bit overly complicated. The problem is to interpolate a 2D array (i.e. age and magnitude), correct? In that case, these two functions can probably be simplified to a few lines of codes by using e.g. `scipy.interpolate.interp2d`, `scipy.interpolate.griddata`, or `scipy.interpolate.LinearNDInterpolator`.

- Perhaps rename the module to `MassLimitsModule`?

tomasstolker:
I have added a test case for the `MassCurveModule` in `test_processing.test_limits`. Note that the use of `urllib` is different in Python 2 and Python 3.

benassa-de-glassa:
@tomasstolker I have changed the dictionary tags and the module name name, as well as adapted the test cases.
I have implemented the interpolation this way, because I got it to work correctly. I will look into the scipy functions you mentioned.

benassa-de-glassa:
Hey @tomasstolker, @18alex96 ,
I've implemented the `scipy.interpolate.griddata` function for the model interpolation.. However it fails the mass_limits test. I can't figure out why. I'd appreciate it if you could have a look at it.
Thanks, Benedikt

tomasstolker:
Probably you need to adjust the results in the test cases since you are using a different interpolation method now. So the values will likely have changed a bit.

To test if the interpolation is correctly implemented, please run it with a few age-magnitude combinations as given in the (discrete) input grid. You should then retrieve the masses as given in the isochrones table.

benassa-de-glassa:
@tomasstolker 

> Probably you need to adjust the results in the test cases since you are using a different interpolation method now. So the values will likely have changed a bit.

I figured that they would have to change, but I'm off by a factor of 10 which I figure is a bit much.

> To test if the interpolation is correctly implemented, please run it with a few age-magnitude combinations as given in the (discrete) input grid. You should then retrieve the masses as given in the isochrones table.

I have done that and do retrieve the masses. I'm confused by the test result.

tomasstolker:
> I have done that and do retrieve the masses. I'm confused by the test result.

OK good! Perhaps there was a bug in the previous implementation of the interpolation...

tomasstolker:
Thanks for updating the module @benassa-de-glassa . I have made some further improvements, please have a look.

Can you add the parameters+types+description in the functions of the module?

One part that I don't fully understand is the `interpolate_model` function. It seems a bit too complex for what it is doing (I think, perhaps I have misunderstood). Is it possible to simply read the column with masses and the column with magnitudes (as function of age) from the isochrones file and create a `numpy.meshgrid` from that as input for `griddata`?

benassa-de-glassa:
> Can you add the parameters+types+description in the functions of the module?

Yes I will get on that first thing tomorrow.

> One part that I don't fully understand is the `interpolate_model` function. It seems a bit too complex for what it is doing (I think, perhaps I have misunderstood). Is it possible to simply read the column with masses and the column with magnitudes (as function of age) from the isochrones file and create a `numpy.meshgrid` from that as input for `griddata`?

I kept having issues with `np.meshgrid` because the ages and contrasts are not given on a regular grid, but the contrast and masses shift over time. I was having issues taking this into account while using meshgrid. 
I therefore decided it would be easier to write a function which returns the mass for a given pair of age and contrast.

tomasstolker:
OK but in that case I would read for each age the column of masses and magnitudes from the grid and add a third column with the (constant) age. This you can do for each age and combine them in 3 long columns.

What I don't like about the current implementation is the use of `numpy.argwhere` on the age and magnitude (which are floats) to find their position in the grid for these values. This seems a bit tricky...

benassa-de-glassa:
Yeah, I've implemented this now. That is indeed way simpler, thanks for the suggestion.

PR #340, opened 2019-04-30
Near initialization module

404NotFoundv2:
Hee Tomas,

Here the pull request for the near module. I (of course) copied quite a bit from the [fitsreading](https://github.com/PynPoint/PynPoint/blob/master/pynpoint/readwrite/fitsreading.py) module, but I remember seeing some pull requests about that module, so it might not be in complete sync.

Lastly, one point of attention are the variables [exp_no_sky/science](https://github.com/PynPoint/PynPoint/blob/ab1c36d2d98b62215b0e3ec9df78d5a86bc381fc/pynpoint/processing/background.py#L472) and [nframes_sky/science](https://github.com/PynPoint/PynPoint/blob/ab1c36d2d98b62215b0e3ec9df78d5a86bc381fc/pynpoint/processing/background.py#L475) in the [NoddingBackgroundModule](https://github.com/PynPoint/PynPoint/blob/ab1c36d2d98b62215b0e3ec9df78d5a86bc381fc/pynpoint/processing/background.py#L393). The subtraction module is ran beforehand to subtract the chops and doesn't update these attributes. Therefore when you run the NoddingBackgroundModule these variables will give twice the amount of frames. I haven't found a proper fix yet, as I didn't put any effort in this yet :grimacing: (And I'm not sure if this would affect other modules)


404NotFoundv2:
I tried to add a test case for this module, but I can't seem to find any documentation about it online :confused: Is this only available when you pay for it?

Anyway, if you're to update the test file this is useful to know:
- the PrimaryHDU only contains 1 big general header, no data
- the next HDU, an ImageHDU contains both a (small) header and data (1 frame)
- if there are 500 images, there is 1 PrimaryHDU and 501 ImageHDUs. The last one is an average of all frames. This is a way they programmed the instrument output. I've never included the last frame, and also this module 'omits' it. But this means that if you want for instance 5 frames as output you need to create 6 imageHDUs in the test cases as it doesn't take the last one into account.

I hope this is not too vague. Otherwise, let me know.

Furthermore, I think the module works pretty well. Maybe we should change the name `near.py`, but I have no good idea.

tomasstolker:
Hi @404NotFoundv2,

The test cases in the `test` folder can be executed - free of charge ;) - with `pytest`. Either from the folder itself or with `make test` from the main folder where the `Makefile` is located. You can install `pytest` for example with `pip`.

I have done a bit of rewriting and restructuring of the NEAR modules, mostly to simplify things and make a few parts more readable. I have also included test cases in `test_nearreading.py`. I think that I understand the data format and the required header keywords but please check if this was done correctly.

Some additional details:

- Renamed the pipeline module to `NearReadingModule` and put it in `readwrite.nearreading`.

- I think that it would be better to always check the attributes since the module is dedicated for a single instrument (in contrast to `FitsReadingModule` which is generic), so I removed the `check` argument.

- Also I removed the `overwrite` argument because it was actually not used anywhere in the code.

- Renamed the output port arguments to something more explicit.

- Created `util.attributes` which includes functions to add the attributes to a dataset based on the configuration file. In this way there is no code duplication between `NearReadingModule` and `FitsReadingModule`.

- Added some more comments and updated the autodocs.

- The nodding position is now inferred from the exposure number (instead of the file number). So if a FITS file is missing it will still provide the correct nod position as long as the first FITS file corresponds to the first position of the chosen nodding scheme.

- Added `util.tests.create_near_data` which creates test data in the VISIR format.

Tested on two FITS files and checked the attributes of the output tags. Also the `StackCubesModule` and `NoddingBackgroundModule` work for me, I think that there was a bug with the `nimages` which had to be divided by 2 in order to have the number of images per chop position.

We should indeed update the documentation page when the `NearReadingModule` is available but (for now) I would leave the example with the test data since the commissioning data is not publicly available (right?), but later on we should update the example based on real data.

Please have another look @404NotFoundv2 and make further changes where needed. Further comments and suggestions are welcome...

404NotFoundv2:
Sounds good to me @tomasstolker !

> The test cases in the test folder can be executed - free of charge ;) - with pytest

Good to know!

> We should indeed update the documentation page when the NearReadingModule is available but (for now) I would leave the example with the test data since the commissioning data is not publicly available (right?), but later on we should update the example based on real data.

I would agree indeed. It would not cost too much effort to edit the pages when the code is written.

I'm happy you found a good way to update the header NAXIS3, as I couldn't :smile:
Furthermore the code is indeed a lot more readable, great!

I do have a few (minor) comments:

- I left the header keyword `ESO SEQ NODPOS` still there to check, because it will always be more reliable than the manual method implemented now. You think it's not worth checking? Or would you add it when the data appears to have this keyword?
- I'm thinking out loud here: If the FITS file contains an odd number of images, you increase nimages with 1. Then you loop over nimages, which will eventually be 1 larger than the actual number of images, right? The only reason why an error will never occur, is that the last image acutally exist (the average over all images) which we do not include. Would it not be better to do `nimages -= 1`?
- I spotted a few minor errors in the comments, which I pushed already.
- For each fits file it spawns 'double information' - of which the variable `DITHER_X` and `DITHER_Y` is also set to `None` in the configuration file of  `docs/near.rst`. So isn't this always happening?:
<pre>Initiating PynPoint v0.7.0... [DONE]
Validating module near... [DONE]
Running NearReadingModule... 0.0% 
Running NearReadingModule... 50.0% 
Running NearReadingModule... [DONE]
/scratch/jjonker/jasper_test_3/lib/python3.6/site-packages/PynPoint/pynpoint/util/attributes.py:119: UserWarning: Non-static attribute DITHER_X (=ESO SEQ CUMOFFSETX) not found in the FITS header.
  'FITS header.' % (attr, fitskey))
/scratch/jjonker/jasper_test_3/lib/python3.6/site-packages/PynPoint/pynpoint/util/attributes.py:119: UserWarning: Non-static attribute DITHER_Y (=ESO SEQ CUMOFFSETY) not found in the FITS header.
  'FITS header.' % (attr, fitskey))
/scratch/jjonker/jasper_test_3/lib/python3.6/site-packages/PynPoint/pynpoint/util/attributes.py:119: UserWarning: Non-static attribute DITHER_X (=ESO SEQ CUMOFFSETX) not found in the FITS header.
  'FITS header.' % (attr, fitskey))
/scratch/jjonker/jasper_test_3/lib/python3.6/site-packages/PynPoint/pynpoint/util/attributes.py:119: UserWarning: Non-static attribute DITHER_Y (=ESO SEQ CUMOFFSETY) not found in the FITS header.
  'FITS header.' % (attr, fitskey))</pre>
- Lastly, when I started writing the code, I also had the size of chopa & chopb as `nimages/2` : 
```python
  # shape of the image stacks for chop A/B (hence nimages/2)
  im_shape = (nimages // 2, header_image['NAXIS2'], header_image['NAXIS1']) 
  .
  .
  .
  # initialize the image arrays for chop A and B
  chopa = np.zeros(im_shape, dtype=np.float32)
  chopb = np.zeros(im_shape, dtype=np.float32)
```

  But this seemed to give an error eventually, which is why I initialized chopa & chopb with the full size `nimages` instead of `nimages/2` and at the end resize. I can't recall why this error happened, so it is a bit vague/not useful right now, but might be good to know later.

Besides that, I tested it on 2 random commisioning fits files and it seemed to work well!

tomasstolker:
> I left the header keyword ESO SEQ NODPOS still there to check, because it will always be more reliable than the manual method implemented now. You think it's not worth checking? Or would you add it when the data appears to have this keyword?

I think that it would be good to check ourselves (again) when new data becomes available, but I would not mention this keyword in the documentation or send a warning to the user (to keep things as simple as possible), since the keyword is most likely not used for NEAR (at least, it was not present in the commissioning data). On the other hand, if we want to make this module more generally available also for archival VISIR data then it might good to check for this keyword. Is that a requirement @spquanz?

> I'm thinking out loud here: If the FITS file contains an odd number of images, you increase nimages with 1. Then you loop over nimages, which will eventually be 1 larger than the actual number of images, right? The only reason why an error will never occur, is that the last image acutally exist (the average over all images) which we do not include. Would it not be better to do nimages -= 1?

Yes I agree! Better have -1. I will change it. With your experience working with VISIR data, did it ever (or commonly) happen that the number of chop A and chop B images was not the same in a FITS file and/or the total number of images was an odd number? We should of course have some checks (as currently is the case) but I am wondering how careful we have to be here.

> For each fits file it spawns 'double information' - of which the variable DITHER_X and DITHER_Y is also set to None in the configuration file of docs/near.rst. So isn't this always happening?

I am not fully sure if I understand you correctly. If you set the DITHER_X and DITHER_Y to None in the configuration file then the warnings should not appear, correct? The reason they appear 4 (?) times is that the attributes have to be written to 4 output ports (nod A/B, chop A/B).

> Besides that, I tested it on 2 random commisioning fits files and it seemed to work well!

OK great! Thanks for testing!

404NotFoundv2:
> With your experience working with VISIR data, did it ever (or commonly) happen that the number of chop A and chop B images was not the same in a FITS file and/or the total number of images was an odd number? We should of course have some checks (as currently is the case) but I am wondering how careful we have to be here.

No it did not commonly occur, it happened once or twice for me (out of many). One of them had a clear reason (telescope abruptly shut down). So I don't think we have to be extremely cautious.

> I am not fully sure if I understand you correctly. If you set the DITHER_X and DITHER_Y to None in the configuration file then the warnings should not appear, correct? 

You are right. I completely overlooked the fact that I hadn't changed my own configuration file. I checked again when I set DITHER_X/Y to none and it does not give me a warning :smiley:

tomasstolker:
All checks are passing. The only issue is the use of `shell=True` in the `subprocess.call` of `util.tests.create_near_data` which is considered a security issue by CodeFactor. I didn't get it to work otherwise on the Travis CI (tried also with `subprocess.check_call`).

PR #341, opened 2019-05-02
Contrast multiprocessing pool

benassa-de-glassa:
I have implemented a worker pool to execute the different `util.contrast_limit` jobs. I started this branch because I got into a weird situation where some Processes stalled and then no results where
retrievable and I could not figure out why. With this implementation I have not yet encountered stalled processes.

I further included a `max_time` parameter, which will ensure that the available results are retrieved after `max_time` (in seconds) has passed. I use this when running a job on EULER, when I want to ensure, that I get as many results as possible in the alloted time.

Additionally, this results in more efficient sequencing of jobs when there are more jobs than processing units: I tried illustrating this below. 

Example `cpu = 4 , len(positions) = 8`.
The `-` symbolize the individual processes working. 
The `.` symbolize a worker being inactive.
The `|` symoblize the overhead from running the program in parallel.

Current:
`|cpu||-----------------------------------------|`
`| 1 ||----1----|........|----5----|............|`
`| 2 ||-----2-----|......|--6--|................|`
`| 3 ||-3-|..............|-----7-----|..........|`
`| 4 ||--------4--------||----------8-----------|`

This pull request:
`|cpu||--------------------------------------|...`
`| 1 ||----1----||--6--||----5----|..........|...`
`| 2 ||-----2-----||-----7-----|.............|...`
`| 3 ||-3-||----5----||---------8-----------||...`
`| 4 ||--------4--------|....................|...`

besides finishing the 8 example processes earlier, it also allows the cpus to start a new process earlier.
Of course the effect is smaller when the jobs take the same time to complete.

I am aware that the sorting towards the end of the module is using my old dictionary method. I have implemented `try/except` statements which should make both methods useable.


tomasstolker:
Very nice work @benassa-de-glassa!

@404NotFoundv2 would you mind having a look and provide some feedback?

> I am aware that the sorting towards the end of the module is using my old dictionary method. I have implemented try/except statements which should make both methods useable.

Why is the dictionary required for you if the processes don't stall anymore? We should not implement features that are only required for local implementations by a user, but think about the bigger picture of PynPoint instead. So, I think that it would be better the fix the origin of the problem by making sure that no processes get stalled in the `ContrastCurveModule`. I never ran into this problem myself, did other users?

404NotFoundv2:
Hee @benassa-de-glassa 
Thanks for the further improvement of the contrast module! 

I'm not sure what you mean by stalled processes? Currently if the cpu is set to 4 and there are more than 4 positions to process, it will start the first 4 positions. Some finish a bit earlier but they wait till all 4 are done. Then it will repeat this. Are these so-called 'stalled processes' simply idle processors doing nothing when it is waiting for any of the other processes to finish? 

If that is the case, it don't see why this is (giving you) a problem?
I do agree it could be improved with a method you describe above by allowing it to start a new job earlier. This will probably decrease the time by about 5-10% 

Lastly, could you elaborate a bit more why we should have a _max_time_ parameter? Aren't we waiting until the result has converged to a value? What is the benefit of adding this parameter? Isn't this a specific user case when running pynpoint on euler? Then you can submit your job on euler with enough hours to finish this module.

benassa-de-glassa:
@tomasstolker 
> Why is the dictionary required for you if the processes don't stall anymore? 

I put in a time limit, which I found very useful when doing the computations on EULER. If it the alloted time is not long enough to go through all positions, it will still go as far as possible and save the result, the computation will not have been "wasted", when nothing was written to the Pypeline later on. It covers the case when not all positions were calculated. If you want to keep it general, there is no problem changing the code. Basically I liked having a kill switch which would jump in and give me some results even when some results had stalled.

> We should not implement features that are only required for local implementations by a user, but think about the bigger picture of PynPoint instead. So, I think that it would be better the fix the origin of the problem by making sure that no processes get stalled in the ContrastCurveModule. I never ran into this problem myself, did other users?

Yes, I agree. I was not able to figure out why they stall. I ran into these stalls for different angles and positions, and it was not perfectly reproducible among identical runs. The `sklearn.decomposition.PCA` function uses a RNG, but I can't imagine it causing differences significan't enough to cause such stalls.

I rather believe that the issue for me was not with the `contrast_limits` function itself. I suspect, and please be aware that my understanding of this is limited, that two `mp.Process` were trying to write to the `mp.Queue` at the same time, leading to the `mp.Queue` object not knowing which result to put first and stalling there.

@404NotFoundv2 
> I'm not sure what you mean by stalled processes? Currently if the cpu is set to 4 and there are more than 4 positions to process, it will start the first 4 positions. Some finish a bit earlier but they wait till all 4 are done. Then it will repeat this. Are these so-called 'stalled processes' simply idle processors doing nothing when it is waiting for any of the other processes to finish?

By stalled processes I mean processes not joining at all while being at 0% cpu usage. I understand that there were supposed to be idle processors while waiting for the i * 4th job to join, but I waited up to 24 hours for one process to join on EULER, while 23 other processing were idling, this can not be the program working as intended.

> If that is the case, it don't see why this is (giving you) a problem?

Since the next set of processes has to wait for all `cpu` number of processes to finish, the entire module stalls at that point and no results are available. 
My best guess for why I encounter stalling processes is elaborated above. 

> I do agree it could be improved with a method you describe above by allowing it to start a new job earlier. This will probably decrease the time by about 5-10%

I have not done a timing comparison. Even in the worst case this way it will not be slower.

> Lastly, could you elaborate a bit more why we should have a _max_time_ parameter? Aren't we waiting until the result has converged to a value? What is the benefit of adding this parameter? Isn't this a specific user case when running pynpoint on euler? Then you can submit your job on euler with enough hours to finish this module.

The max_time parameter is a convenience parameter, this way one does not have to wonder if the submitted time is sufficient for all positions, but instead one will get as many results as possible in that time and warnings for those which were not calculated.

tomasstolker:
> I put in a time limit, which I found very useful when doing the computations on EULER. If it the alloted time is not long enough to go through all positions, it will still go as far as possible and save the result, the computation will not have been "wasted", when nothing was written to the Pypeline later on.

Since the processing time for each position is approximately the same, I would simply calculate a few positions first and then upscale the time to the total number of positions that you want to process. If you want to calculate limits from very large image stacks (I assume that this is what you are doing) you could also subdivide the separation range of the `ContrastCurveModule` in a for loop. So I don't think that we have to build in a time limit as it seems an EULER-specific problem.

> I rather believe that the issue for me was not with the contrast_limits function itself. I suspect, and please be aware that my understanding of this is limited, that two mp.Process were trying to write to the mp.Queue at the same time, leading to the mp.Queue object not knowing which result to put first and stalling there.

@404NotFoundv2 @majobodo @timothygebhard Any thoughts on this?

> By stalled processes I mean processes not joining at all while being at 0% cpu usage. I understand that there were supposed to be idle processors while waiting for the i * 4th job to join, but I waited up to 24 hours for one process to join on EULER, while 23 other processing were idling, this can not be the program working as intended.

Is this a problem you only have on EULER or also on other machines?

benassa-de-glassa:
> Is this a problem you only have on EULER or also on other machines?
It also occured on bluesky. But I don't have the traceback from it.

404NotFoundv2:
> I rather believe that the issue for me was not with the contrast_limits function itself. I suspect, and please be aware that my understanding of this is limited, that two mp.Process were trying to write to the mp.Queue at the same time, leading to the mp.Queue object not knowing which result to put first and stalling there.

I don't think this is the case as queues are made for this. They are [thread and/or process safe](https://docs.python.org/3.7/library/queue.html#). This is the reason why I implemented Queues instead of a return statement.

Without a the traceback/log code, it is guessing where the problem lies. 
The last thing I can still think of why this single process took so long is because it didn't converge, while the others did. I remember somewhere that pynpoint used to keep trying 50 times to see if it converges. I don't know if this is still the case. @tomasstolker will know better.

tomasstolker:
> I don't think this is the case as queues are made for this. They are thread and/or process safe. This is the reason why I implemented Queues instead of a return statement.

Ok good, that is reassuring :).

> The last thing I can still think of why this single process took so long is because it didn't converge, while the others did. I remember somewhere that pynpoint used to keep trying 50 times to see if it converges. I don't know if this is still the case. @tomasstolker will know better.

The new implementation of the `ContrastCurveModule` requires a single PSF subtraction per position and not an iteration so that can not be the problem. I assume that you are using this new implementation @benassa-de-glassa?

timothygebhard:
> I don't think this is the case as queues are made for this. They are thread and/or process safe. This is the reason why I implemented Queues instead of a return statement.

I agree, the whole point of using queues from the `multiprocessing` module is to ensure thread-safety.

*However*, in my experience, queues can be a bit tricky to work with (and don't always behave the same on different operating systems). For example, if you have multiple processes that put their results to into a mutual `result_queue`, you might have to make sure that you empty that queue regularly because otherwise the worker processes don't finish properly and cause a deadlock. See [the Python docs](https://docs.python.org/3/library/multiprocessing.html#pipes-and-queues):

> **Warning:** As mentioned above, if a child process has put items on a queue (and it has not used `JoinableQueue.cancel_join_thread`), then that process will not terminate until all buffered items have been flushed to the pipe.
This means that if you try joining that process you may get a deadlock unless you are sure that all items which have been put on the queue have been consumed. Similarly, if the child process is non-daemonic then the parent process may hang on exit when it tries to join all its non-daemonic children.
Note that a queue created using a manager does not have this issue. See Programming guidelines.

If I understand correctly, though, the point of this PR was to *remove* the explicit usage of queues and instead use a `Pool` of workers and `apply_async`, with the idea that this would be faster because things could be processed asynchronously? 

I've never used `apply_async` before, but I have written my own "multiprocessing manager" to use multiple processes to asynchronously work through an argument queue. I think that would be closer to the current implementation, except that processes don't wait for each other. If there's any interest, I'd be happy to share the code! (@tomasstolker already knows what I'm talking about.)

404NotFoundv2:
> I've never used apply_async before, but I have written my own "multiprocessing manager" to use multiple processes to asynchronously work through an argument queue. I think that would be closer to the current implementation, except that processes don't wait for each other. If there's any interest, I'd be happy to share the code! (@tomasstolker already knows what I'm talking about.)

I think it is always good to see more than one option. It would be nice if you could share it (if that doesn't take too long).

tomasstolker:
> However, in my experience, queues can be a bit tricky to work with (and don't always behave the same on different operating systems). For example, if you have multiple processes that put their results to into a mutual result_queue, you might have to make sure that you empty that queue regularly because otherwise the worker processes don't finish properly and cause a deadlock.

I see, perhaps that is causing the problem, because in the current `ContrastCurveModule` the queue is only emptied once, that is, in the end when all the processes (should have) finished and joined.

How many angles and separations were you processing when the error occurred @benassa-de-glassa?

benassa-de-glassa:
> The new implementation of the `ContrastCurveModule` requires a single PSF subtraction per position and not an iteration so that can not be the problem. I assume that you are using this new implementation @benassa-de-glassa?

Yes I am using the current implementation.

> How many angles and separations were you processing when the error occurred @benassa-de-glassa?

I had 40 separations at 8 angles. 
I now know that this is way more than necessary.

Also, I thougt I had appended the Euler traceback in an earlier comment but must have forgotten. I am sorry. I do not have a traceback from bluesky. I killed this particular job using `bkill`.
[euler_output.txt](https://github.com/PynPoint/PynPoint/files/3141329/euler_output.txt)

tomasstolker:
Ok let's try to conclude on this.

To make sure I have not misunderstood anything...

When using a `Pool` instead of a `Queue`, there haven't been any stalling processes? Did you test this with the same parameter values (the total number of positions in particular)?

The `max_time` parameter was only added for your convenience in case the job too longer than the requested time?

timothygebhard:
> I think it is always good to see more than one option. It would be nice if you could share it (if that doesn't take too long).

No, [it's already on GitHub](https://github.com/timothygebhard/ggwd/blob/master/generate_sample.py#L240) :) The key idea is that you have the main process run in a loop until you are done processing everything, and the main process is responsible for spawning worker processes, checking their runtime (and aborting jobs that take too long), and removing things from the results queue. Admittedly, it's not perfectly asynchronous, but maybe it can still serve as an inspiration for the problem at hand?
Minor comment: Besides not really giving a reliable number, `queue.qsize()` will cause trouble on macOS, so maybe try to avoid it if you can.

tomasstolker:
I removed the `max_time` parameter and the dictionary which would store partial results. So the only change compared to the previous `ContrastCurveModule` is the use of a multiprocessing `Pool` instead of a `Queue` which is more robust if a very large number of positions are processed.

I changed the waiting time to test how many processes have finished from 60 to 5 seconds. Otherwise for a small dataset (e.g. 100 images) the runtime is set by the waiting time...

Is the `except mp.TimeoutError` still required @benassa-de-glassa?

Could you check the changes one more time before we merge it @benassa-de-glassa? Thanks.

benassa-de-glassa:
> Is the `except mp.TimeoutError` still required @benassa-de-glassa?

The `except mp.TimeoutError` is there to catch the `async_result.get(timeout=0)`. In case the result is not gettable in the instance the `.get`-function is called the exception would still be raised. We can just remove the `timeout=0` parameter.

> Could you check the changes one more time before we merge it @benassa-de-glassa? Thanks.

I ran it for one of my datasets and it looks correct.

tomasstolker:
OK thanks! As we just discussed, I have removed the try-except and `timeout=0`.

PR #342, opened 2019-05-02
Multiprocessing functionalities for subsets of images

tomasstolker:
I have added multiprocessing functionalities for subsets of images in `util.multistack`. The classes inherit from the abstract interfaces in `util.multiproc` (i.e. use the generic master-worker scheme with the poison pill method). There is a `StackReader` which reads in subsets of images, a `StackTaskProcessor` which applies a function on each image of a stack, and a `StackProcessingCapsule` which creates and runs the processes.

The `StackProcessingCapsule` is used by the `apply_function_to_images` function such that any pipeline module using that function (currently there are 17 such modules), will process stacks of images in parallel if `CPU > 1` in the config file. The number of images per stack is `MEMORY` divided by `CPU`. If `CPU = 1` then images are processed 1 at a time by the `apply_function_to_images`. If `MEMORY = None` or the input and output tag are the same, then all images are read into the memory at once and processed 1 at a time (this is only possible if the shape of the image stack does not change).

I have added test cases for all pipeline modules that use the `apply_function_to_images` which run both with `CPU = 1` and `CPU = 4` and check that the results are the same.

Some of the output of the pipeline modules can not be used if `CPU > 1`, for example if there are additional output ports (e.g. `mask_out_tag` in `StarCenteringModule` and `index_out_tag` in `StarExtractionModule`) or printing the image index that causes an error in the `StarExtractionModule`. So such functionalities are only used with `CPU = 1`.


tomasstolker:
One more thing I had forgotten to mention. Before, `util.multiproc` would only allow for multiprocessing of stacks of images for which the output was again a stack of images (i.e. the number of dimensions had to be conserved). I have updated this such that the number of dimensions can change, as long as the first dimension remains equal to the total number of images (otherwise the `apply_function_to_images` can not be used anyway).

PR #343, opened 2019-05-02
Dropping compatibility with Python 2.7

tomasstolker:
I have dropped the compatibility with Python 2.7. There were too many small issues/challenges to keep supporting both Python 2 and Python 3. Also updating of the dependencies was often no longer possible because some of the dependencies have already dropped their support for Python 2.

Some details:

- Removed the `from __future__ import ...` statements.

- Updated the `dims` (being deprecated) to the `shape` argument in `numpy.unravel_index`.

- Removed `(object)` from all classes.

- Updated all the dependency versions in the `requirements.txt`.

- Catched the `FutureWarning` in the test cases of the `MCMCsamplingModule`.

- @18alex96, I had to adjust the test case results of the `AngleCalculationModule` after updating `astropy` from version 2 to 3. Only at the 7th decimal or so though...


18alex96:
> @18alex96, I had to adjust the test case results of the AngleCalculationModule after updating astropy from version 2 to 3. Only at the 7th decimal or so though...

Sure, that's absolutely reasonable and fine with me.

tomasstolker:
- Removed `six` from the `requirements.txt`.

- Updated `six` statements (used for compatibility with both Python 2 and 3) to code that is only compatible with Python 3.

- Updated the configuration of Travis CI, only testing Python 3.6 and 3.7.

timothygebhard:
Does that also mean we will / want to start using [f-strings (PEP 498)](https://www.python.org/dev/peps/pep-0498/) and [type hints (PEP 484)](https://www.python.org/dev/peps/pep-0484/) now? :)

tomasstolker:
@timothygebhard yes sounds good! That seems quite useful for enhancing the readability of the code (I didn't use f-strings and type hints myself so far...), so let's use it for new code (and feel free the update existing code).

While reading about type hints, I came across the [`typeguard`](https://pypi.org/project/typeguard/) package. By adding a decorator to a function it will check during runtime if the argument and return types are correct. This seems very helpful for bug fixing but in particular also for ensuring that users provide the correct types to the pipeline module's arguments. Currently types are being checked here and there in the pipeline modules but it was too much work to implement consistently throughout PynPoint. So I want to propose to add the `@typechecked` decorator to all functions that have been updated with type hints, such that types are checked automatically.

tomasstolker:
Goodbye Python 2! Thanks for all the fun we had...

PR #344, opened 2019-05-07
output from StarCenteringModule in ShiftImagesModule

tomasstolker:
Alternative use of the `shift_xy` parameter in `ShiftImagesModule` which can be a tag name with the fit results of the `StarCenteringModule`. This can be useful if the fit and actual centering should occur at different points in the data reduction. For example, the centroids are fitted before the background subtraction and the shifting after the background subtraction. This can be used for NACO AGPM data (perhaps also for NEAR?).


tomasstolker:
The units of the shift in x and y direction in the best-fit results that are written by the `StarCenteringModule` (`fit_out_tag`) have been changed from arcsec to pixels.

PR #346, opened 2019-05-09
Progress bar tells ETR

benassa-de-glassa:
hey @tomasstolker,

as I just told you the `progress` now _can_ tell you how much time approximately remains until a module has completed. 
To do so, call the function as `progress(current, total, message, start_time)` with `start_time=time.time()`. `start_time` has to be initialized once before the loop.


tomasstolker:
Very nice! I like it :).

Before we merge it, could you check the files that you changed with `pylint` for easy style improvements such as line length, white spaces, etc? Thanks.

PR #358, opened 2019-05-21
FrameSimilarityModule and RemoveFramesByAttributeModule

benassa-de-glassa:
Hey all,

I wrote these two modules, maybe someone else has some use for them. I aim to reproduce the three similarity measures mentioned in this paper: 
[https://iopscience.iop.org/article/10.3847/1538-3881/aafee2/pdf](url)
There is a fourth measure called 'DSC' (Dice Soerensen Coefficient) which doesn't currently give useful results and is commented out. I will update it, once I get it to work.
The SSIM measure does not exactly work as described in the paper above, explicitly it looks at the entire frame instead of a mask over which the window moves. This is because I use the `sklearn.measure.compare_ssim` function which is magnitudes faster than my own implementation.

The Modules are meant to work in synergy, i.e. first calculating the similarity and then removing the least similar frames.

There is also a basic implementation of a jackknife and a bootstrap estimator wrappers in `/util/resampling.py`. These come in handy when finding the similarity large dataset, but are currently not native to the FrameSimilarityModule.

Best, Bene


tomasstolker:
Cool! Thanks a lot for the PR!

- If possible, could you create a couple of test cases? See the `test_processing` folder for some examples.

- Please have a look at the issues that were found by CodeFactor (and do a quick check with pylint if not done yet).

- The URL to the paper seems incorrect in this PR.

benassa-de-glassa:
- I'll work on the test cases tomorrow
 - I changed the issues from codefactor 
   -regarding pylint: I used the variable names due to the resemblance to the cited paper.. should I still rename the variables to be snake_case like?
- I fixed the url.. I don't know why it was linking to another PR...

benassa-de-glassa:
@tomasstolker 
I added four test cases. Writing them felt more difficult then writing the actual module :D. 
Let me know if you think I missed some testing property.

tomasstolker:
Nice work, thanks a lot for adding the unit tests!

I have a few comments and questions about the modules and util functions:

- The two util functions are not being used by the modules that you added, correct? What is their purpose? Also, there are probably ready-to-use numpy/scipy/astropy functions for jackknife and bootstrap estimators? So I am not sure if these util functions are really required.

- The description of the `FWHM` argument is missing.

- What is special about the covariance and standard deviation functions in `_similarity`? Can we not use numpy functions directly?

- Please remove the not yet working code that is commented out and create a new PR if you decided to implement this.

- Could you add the modules to the list of `docs/overview.rst`?

- I would store the similarity values as dataset instead of attribute. This will simply things also a bit with the implementation because as attribute we otherwise have to make sure that the list gets updated whenever frames are removed. And the values would in general not be valid anymore after an (additional) frame selection since the median reference frame has changed.

- The `run` method of `RemoveFramesByAttributeModule` needs some more details for the autodocs.

- Perhaps rename `RemoveFramesByAttributeModule` to `SelectSimilarFrames`?

benassa-de-glassa:
following your email I added lines 894 and 944 which close and reopen the output port while it is not required to be open

benassa-de-glassa:
> Nice work, thanks a lot for adding the unit tests!
> 
> I have a few comments and questions about the modules and util functions:
> 
>     * The two util functions are not being used by the modules that you added, correct? What is their purpose? Also, there are probably ready-to-use numpy/scipy/astropy functions for jackknife and bootstrap estimators? So I am not sure if these util functions are really required.

Yeah, you're right. I did not even think of using the numpy/scipy/astropy functions, because I still had these functions from last years statistic course... 
I figured that if somebody else wanted to use them when writing a new module they might find it helpful that this resampling method already exists.
In this specific context, their purpose is that I wanted to open up a discussion about them. Basically when there is a huge dataset of images you want to compare to a common temporal median, I thought it might be feasible to use a subset of them to compute the temporal median. I opted to use the `temporal_median` parameter for now which calculates the temporal median once, but wanted to leave the option open to use resampling.

I will remove them. It makes no sense having them in when there is a scipy implementation.

>     * The description of the `FWHM` argument is missing.

ups... sorry, I'll get on it.

>     * What is special about the covariance and standard deviation functions in `_similarity`? Can we not use numpy functions directly?

I had issues with the shape and dimensionality of the images, compared to the source paper. But I think I figured it out thanks to Anna. I would appreciate it if you double checked.

>     * Please remove the not yet working code that is commented out and create a new PR if you decided to implement this.

Will do.

>     * Could you add the modules to the list of `docs/overview.rst`?

Will do.

>     * I would store the similarity values as dataset instead of attribute. This will simply things also a bit with the implementation because as attribute we otherwise have to make sure that the list gets updated whenever frames are removed. And the values would in general not be valid anymore after an (additional) frame selection since the median reference frame has changed.

So you mean to use a separate tag, with its own attributes etc? 
I feel like this would complicate things.. 
 - As it currently is, the attribute stick to each frame so one has a one-to-one comparison. 
 - In the current pynpoint implementation the FrameSelectionModule updates the attrbutes correctly already when removing frames.
 - The fact that they have to be recalculated when reselecting frames is still true also when using a separate tag. 



>     * The `run` method of `RemoveFramesByAttributeModule` needs some more details for the autodocs.

Will do.

>     * Perhaps rename `RemoveFramesByAttributeModule` to `SelectSimilarFrames`?
I named it this way, because I've written the module in a more general fashion such that one could also select Frames using any attribute a tag has. If you wanted to you could also use it to select the first 100 frames using the INDEX attribute tag.

tomasstolker:
> > * What is special about the covariance and standard deviation functions in `_similarity`? Can we not use numpy functions directly?
> 
> I had issues with the shape and dimensionality of the images, compared to the source paper. But I think I figured it out thanks to Anna. I would appreciate it if you double checked.

@aboehle would you mind checking? I am not familiar with the details...

> * As it currently is, the attribute stick to each frame so one has a one-to-one comparison.
> * In the current pynpoint implementation the FrameSelectionModule updates the attrbutes correctly already when removing frames.

Ah right, I once implemented that, good :).

> * The fact that they have to be recalculated when reselecting frames is still true also when using a separate tag.

OK let's make it an attribute then.

> > ```
> > * Perhaps rename `RemoveFramesByAttributeModule` to `SelectSimilarFrames`?
> > ```
> 
> I named it this way, because I've written the module in a more general fashion such that one could also select Frames using any attribute a tag has. If you wanted to you could also use it to select the first 100 frames using the INDEX attribute tag.

I don't fully understand how it works. It selects the frames by the highest attribute value?

benassa-de-glassa:
> > > * What is special about the covariance and standard deviation functions in `_similarity`? Can we not use numpy functions directly?
> > 
> > 
> > I had issues with the shape and dimensionality of the images, compared to the source paper. But I think I figured it out thanks to Anna. I would appreciate it if you double checked.
> 
> @aboehle would you mind checking? I am not familiar with the details...

The implementation passed the test I wrote yesterday, before using the numpy function. So I think it is fine, but I too would be super happy if it were double checked.

> > I named it this way, because I've written the module in a more general fashion such that one could also select Frames using any attribute a tag has. If you wanted to you could also use it to select the first 100 frames using the INDEX attribute tag.
> 
> I don't fully understand how it works. It selects the frames by the highest attribute value?

Pretty much, you can also make it the lowest attribute value if you were interested in those be making the `order` parameter `order=ascending`.
I am also planning on allowing the module to be passed specific indices, this way one could cherry pick specific frames.

tomasstolker:
> Pretty much, you can also make it the lowest attribute value if you were interested in those be making the `order` parameter `order=ascending`.
> I am also planning on allowing the module to be passed specific indices, this way one could cherry pick specific frames.

I think that it is important to first decide in which cases these additional implementations are useful (which is currently not clear to me) before making it unnecessarily detailed/complex. Removing frames by index is already possible with another frame selection module.

benassa-de-glassa:
ohh, you're right, I had not seen it. In this case I won't write add the option to pass indices.

aboehle:
> > @aboehle would you mind checking? I am not familiar with the details...
> 
> The implementation passed the test I wrote yesterday, before using the numpy function. So I think it is fine, but I too would be super happy if it were double checked.
> 

I checked the PCC calculation and it is correct! For the record, there is also a scipy.stats.pearsonr function that gives consistent results with your implementation @benassa-de-glassa, but also does extra some calculations to get the p-value that are not necessary for the purposes of this module. So I would stick with the implementation as it is.

benassa-de-glassa:
@aboehle Thanks for checking.

I just noticed a mistake in the previous commit. I multiplied by the pixscale, but the crop_image functions takes the size parameter in pixels, not in arcsec. So there was no need for that.

tomasstolker:
I went through the code and made some changes here and there, also changed some of the parameter and module names and added type checks. Please have a look.

- The `fwhm` parameter is not so clearly explained. It is the FWHM of what? And what do you mean with "relevant scales"?

- The autodocs of the run method of `FrameSimilarityModule` needs some details on the similarity measures that are used, or at least the measures (MSE, PCC, SSIM) should be spelled out once.

- Why is the `INDEX` required by `SelectByAttributeModule` (I renamed the module to something shorter)? Just for me to understand the code a bit better...

benassa-de-glassa:
> I went through the code and made some changes here and there, also changed some of the parameter and module names and added type checks. Please have a look.
> 
>     * The `fwhm` parameter is not so clearly explained. It is the FWHM of what? And what do you mean with "relevant scales"?

The FWHM of the PSF. The relevant scales was incorrect to be put in (I  misremembered the paper). The FWHM appears to be chosen as an arbitrary scale in the Ruane et al. paper.

>     * The autodocs of the run method of `FrameSimilarityModule` needs some details on the similarity measures that are used, or at least the measures (MSE, PCC, SSIM) should be spelled out once.
I added the measures in the docstring.
I'm not really sure what to explicitly write about each measure individually, because the paper which introduced them just tried them out to see how they compare. There is not given any (and I can't think of) a case when one measure is obviously better than the others. Rather one can try them all and then compare their distributions to see if any one of them has some outliers, which might be interesting. 
I'll come by your office later so we can discuss this, ok?

>     * Why is the `INDEX` required by `SelectByAttributeModule` (I renamed the module to something shorter)? Just for me to understand the code a bit better...

Do you mean the example I put in in lines 988-994? 
That is  meant to be an example to show how the Module can be used to order and select for any attribute. You could also do the same using for example `PARANG`(I don't know whether you would want to do that or not) or any other attribute which is available for every frame. For example I also added the coherence time and seeing as an attribute to my personal pipelines, and could use those measure to select frames.

Or do you mean in in line 1062? 
The `write_selected_data` function requires the indices of the images which should be removed. I first order the indices which I want to be used and then cut them at `self.m_number_frames`. 

I hope this helps :D

tomasstolker:
> I'll come by your office later so we can discuss this, ok?

Yes, thanks, let's briefly discuss the final details.

PR #363, opened 2019-06-13
changed ShiftImages, now uses SubtractImages like style for iteration

benassa-de-glassa:
As discussed yesterday, I've implemented the `ShiftImagesModule` into a `SubtractImagesModule.run` type iteration. Additionally I  modified the `util.image.shift_image` function in such a way that it can take 3D stacks of images. 

 - When a input_port with individual shifts is provided, the iteration can't loop over stacks of images, as `scipy.ndimage.interpolation.shift` does not support different shifts within an axis.

- The tests for `test_fit_center_full` and `test_fit_center_mean` fail, but the final `test_shift_images_tag` which relies on these tags passes. Interestingly, for me they also fail on the master branch.

### DIFF
--- a/pynpoint/processing/centering.py
+++ b/pynpoint/processing/centering.py
@@ -1063,18 +1063,46 @@ class ShiftImagesModule(ProcessingModule):
             None
         """
 
+        # delete all data stored in self.m_image_out_port
+        self.m_image_out_port.del_all_attributes()
+        self.m_image_out_port.del_all_data()
+
+        # grab the fit results from the self.m_fit_in_port if available
+        # since each frame is shifted individually, it can not be easily applied to
+        # stacks of frames
         if self.m_fit_in_port is not None:
             self.m_shift = -1.*self.m_fit_in_port[:, [0, 2]] # (x, y)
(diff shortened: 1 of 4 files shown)


tomasstolker:
Thanks!

> When a input_port with individual shifts is provided, the iteration can't loop over stacks of images, as scipy.ndimage.interpolation.shift does not support different shifts within an axis.

There is also the case that the `FitCenterModule` outputs fit values for the mean, in which case the shift values in x and y direction are the same for all images. Perhaps we can build in a check if this is the case such that the `shift` function can still be applied on a 3D array?

> The tests for test_fit_center_full and test_fit_center_mean fail, but the final test_shift_images_tag which relies on these tags passes. Interestingly, for me they also fail on the master branch.

Hm that is weird, the CI seems to have finished correctly on the master and also locally it seems to work for me (but I also get an error on the PR branch).

The centering tests use the output of the shift module, perhaps something went wrong there?

tomasstolker:
Is the transpose required in the `shift_image` function? Perhaps it also works with `vstack` or `hstack`?

PR #368, opened 2019-06-17
Refactor CropImagesModule

timothygebhard:
As discussed with @tomasstolker, I've slightly refactored the `CropImagesModule` such that it doesn't use the `apply_function_to_images()`-based parallelization anymore (which does not make a lot of sense here), but instead process the images in chunks and make use of numpy's vectorization features.

For testing purposes, I ran the module on a data set with 29681 frames. The old version took 960.2 seconds to crop the images, whereas the new version finished in just 29.8 seconds. This corresponds to a speed-up factor of ~32.

I also changed a few things in the `utils/image.py` in order to improve PEP8 compliance and make my linter stop complaining :)

@404NotFoundv2 I only realized when I was almost done with these changes that we had kinda assigned this task to you at last Wednesday's meeting 🙈I hope this is okay for you!

### DIFF
--- a/pynpoint/processing/resizing.py
+++ b/pynpoint/processing/resizing.py
@@ -3,7 +3,8 @@ Pipeline modules for resizing of images.
 """
 
 import math
-import warnings
+import sys
+import time
 
 from typing import Union, Tuple
 
@@ -13,6 +14,7 @@ from typeguard import typechecked
 
 from pynpoint.core.processing import ProcessingModule
 from pynpoint.util.image import crop_image, scale_image
+from pynpoint.util.module import progress, memory_frames
 
 
 class CropImagesModule(ProcessingModule):
@@ -61,7 +63,7 @@ class CropImagesModule(ProcessingModule):
         self.m_center = center
 
         if self.m_center is not None:
-            self.m_center = (self.m_center[1], self.m_center[0]) # (y, x)
+            self.m_center = (self.m_center[1], self.m_center[0])  # (y, x)
 
     @typechecked
     def run(self) -> None:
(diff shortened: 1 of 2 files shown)


tomasstolker:
Excellent! Thanks a lot @timothygebhard!

- Completely removing the `copy` in `crop_image` was not possible? Because of how it was used by some other modules?

- I am not sure if we should replace the `elif` conditions with `else`. Although from a syntax perspective it makes indeed sense to use `else`, I find the `elif` better for readability of the code because it shows clearly the differences between the different cases. Perhaps add an `else` condition with an error instead?

timothygebhard:
> Completely removing the copy in crop_image was not possible? Because of how it was used by some other modules?

Exactly. Trying to remove the copying entirely causes the `test_waffle_center_odd` and `test_waffle_center_even` tests from `test_processing/test_centering.py` to fail, due to something that happens in the `WaffleCenteringModule`. 
If you know what's happening in there, feel free to change it so that we can drop the `copy` argument here, or at least change the default to False :)

> I am not sure if we should replace the elif conditions with else. Although from a syntax perspective it makes indeed sense to use else, I find the elif better for readability of the code because it shows clearly the differences between the different cases. Perhaps add an else condition with an error instead?

Agreed. I reverted this change and added a `RuntimeError` for the "impossible" fifth case that my linter was still complaining about ("local variable 'center' might be referenced before assignment").

tomasstolker:
> Exactly. Trying to remove the copying entirely causes the test_waffle_center_odd and test_waffle_center_even tests from test_processing/test_centering.py to fail, due to something that happens in the WaffleCenteringModule. If you know what's going on in there, feel free to change it so that we can drop the copy argument here, or at least change the default to False :)

Ok thanks, that is indeed worth looking into. I am not sure why the copy would be required for `WaffleCenterinModule`. Perhaps @18alex96 knows. This could be changed in a new PR.

Issue #356, opened 2019-05-20
Multiprocessing bug


tomasstolker:
There is bug in the multiprocessing tools of `util.multistack`. Somehow every 1000 images or so that are written, 8 of them are only partially stored for unclear reasons (independent of number of processes). This does not create an error or warning but it is visible in the images since part of them have values of zero. It is not clear to me if the problem is in the `multistack` utilities or at a higher level in `util.multiproc` (in which case it may also affects the multiprocessing in time). As a quick fix, I will disable the multiprocessing in `apply_function_to_images`.

tomasstolker:
Multiprocessing of stacks of images with `apply_function_to_images` has been disabled with PR #357.

tomasstolker:
I did an additional test in which in created a FITS cube of 1000 x 1000 x 10000 images (i.e. similar in size to the real data that I used before) with some random values. I ran the same procedure again (`RemoveLinesModule`) with and without the multiprocessing and tested the output of each image with `numpy.allclose`. Interestingly/surprisingly the problem did not occur in this case... That is, the output of the images was the same both with and without multiprocessing. I have no clue what this means... I also tested again on the real data, and the problem is still there...

With the real data, this is what an image looks like that was not properly stored (<= 1% of the data). Left is without multiprocessing and right is with multiprocessing:

![Screenshot 2019-05-21 at 08 57 53](https://user-images.githubusercontent.com/3204762/58101899-aef9ad80-7be0-11e9-8f90-e6acb30773f0.png)

Suggestions very much welcome...

tomasstolker:
I have tested from the ``multiproc`` branch with a NACO dataset and have the same issue (as described above) both on Linux and macOS.

benassa-de-glassa:
I can recreate the issue. But the differences occur for different frames than @tomasstolker described... For me those are: `[ 315  316  317  318  319  320  321  322  323  365  366  455  456 1485 1486 1487 1488 1489 1490 1491 1492 1493 1575 1576 1577 1578 1579 1580
 1581 1582 1583]`

![image](https://user-images.githubusercontent.com/33694956/59510794-e6443d00-8eb4-11e9-94a0-ed2d3e479d91.png)

sdobler:
I tested the RemoveLinesModule on my NACO data and got an issue in the multiprocessing about every 7000 images. As for the others the lower section gets zero values.

/home/siri/Pictures/Issue_mp.png

aboehle:
[...]

PR #370, opened 2019-06-18
Memory intensive hotfix for issue #356

benassa-de-glassa:
TaskReader: 
 * instead of accessing the port each time and running into the issue #356, the data is read in once. This obviously is not very feasible for large chunks of data.


tomasstolker:
Thanks for all the efforts...

Since the bug is not affecting the master branch, I don't think that this hotfix is required for the multiproc branch. The mutex is used at other parts of the multiprocessing as well, so we should first understand the origin of the problem and find a general solution.

I will close this PR for now but please reopen if you don't agree with this approach.

PR #372, opened 2019-06-20
Multiprocessing bug fix

tomasstolker:
Bug fix for issue #356.

The problem was related to the database that was shared between processes. Even though the mutex made sure that reading and writing was only possible for one process at a time, it is not possible to share input/output ports (which contain an opened database) between processes.

This is solved by closing all ports in `apply_function_in_time` and `apply_function_to_images` before the multiprocessing starts, and opening and closing the ports (and database) each time the mutex is used.

- The use of `util.multistack` in `apply_function_to_images` has been reactivated.

- Note that the bug was also present `util.multiline` which is for example used by the `WaveletTimeDenoisingModule`.

- For the multiprocessing of the PCA residuals I never experienced any problems. Perhaps the reason is that the multiprocessing is only used for writing of the residuals while the input data is passed directly. Nonetheless, I have also changed the implementation there, that is, the database is opened and closed each time the mutex is used for storing of data.

- I changed the implementation of `RemoveLinesModule` temporarily back since in the master it was no longer using the `apply_function_to_images` while there is a test case for the multiprocessing.


aboehle:
This fix passes the test case on my laptop and on bluesky, also when I adjust the CPU/MEMORY values. I also tested on a NACO data set with CPU/MEMORY values that reproduced the original bug, and the results with and without multiprocessing are the same.

Thanks @tomasstolker!

benassa-de-glassa:
@tomasstolker The tests pass for me too! Thank you!

tomasstolker:
> Will you revert some of the changes from #371 or is the overhead from multiprocessing too high for those?

I never tested the difference in runtime but I would guess that using the vectorized approach with a single process is still faster than using the the multiprocessing in `apply_function_to_images`. The multiprocessing with `apply_function_to_images` is perhaps mainly useful for slow functions such as in `FitCenterModule` when fitting all images separately.

PR #373, opened 2019-06-20
Add chop subtraction and cropping options to NEARReadingModule

aboehle:
Added options to the NEARReadingModule to perform the chop subtraction (chop A - chop B and vice versa) and/or the image cropping (cropping images around a fixed location, i.e., the coronagraph position) prior to writing out the chopA and chopB data sets to the database. The goal with this change is to skip saving some of the intermediate steps of the data reduction. Note that with the default cropping size you do remove the region with the non-coronagraphic PSF.

### DIFF
--- a/pynpoint/readwrite/nearreading.py
+++ b/pynpoint/readwrite/nearreading.py
@@ -43,7 +43,7 @@ class NearReadingModule(ReadingModule):
                  chopa_out_tag: str = 'chopa',
                  chopb_out_tag: str = 'chopb',
                  do_chop_sub: bool = False,
-                 do_chopa_crop: bool = False,
+                 do_crop: bool = False,
                  crop_size: float = 5.,
                  crop_center: Union[tuple, None] = (432, 287)):
         """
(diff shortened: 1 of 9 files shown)


pynpoint/readwrite/nearreading.py
```diff
@@ -365,6 +392,20 @@ def run(self) -> None:
             # get the images of chop A and chop B
             chopa, chopb = self.read_images(filename, im_shape)
 
+            if self.m_do_chop_sub:
+                chopa = chopa - chopb
```
benassa-de-glassa: In the `SubtractImageModule` there is a scaling factor applied. Is that not necessary here?

aboehle: That's correct, the scaling factor isn't necessary for the chop subtraction, so I think fixing that parameter to 1 in the `NEARReadingModule` is ok.

tomasstolker:
Great, thanks! Perhaps we can also add an option to stack the images per FITS cube? By passing it 'mean', 'median' or None? If you have time, could you add a test case that uses these new parameters?

aboehle:
The cube stacking feature is now included! This also required changing the NFRAMES attribute to 1 as the files were read in and adjusting the AngleInterpolationModule to deal with this case. For the commissioning data, the resulting PARANG values are the same as when each step is performed by the separate modules.

I will try to add test cases today! But I may not have time until I'm back after holidays next week.

tomasstolker:
No worries, I have quickly added a test case which uses all new parameters. I have also reduced the number of new parameters to 3.

With the chop subtraction, I have added a `numpy.copy` because I was not sure about `chopb = -chopa`. Perhaps something to test at a later stage, because the copy will require more memory and computation time.

Have a nice holiday!

PR #374, opened 2019-06-21
New features to the contrast curve

benassa-de-glassa:
I added a couple of features to the `ContrastCurveModule`.

- Several tags can be read into the module at once.
- These residuals of these tags can be computed of all tags at once or for each tag individually.
- Added the 'weighted' and 'clipped' method of combining residuals. While advertised in the docstring, they were not functional.

P.S. sorry for the "temp commits", I hadn't figured out the git stash feature.


tomasstolker:
Thanks for the PR.

Could you provide a bit more information about what the goal of the new features are and how they have been implemented?

The code has become quite long and complex, perhaps move some of the features to a separate util function? The `SimplexMinimizationModule` and `MCMCsamplingModule` may also need the features for the 'weighted' and 'clipped' residuals.

benassa-de-glassa:
> Could you provide a bit more information about what the goal of the new features are and how they have been implemented?

The PR implements the features I showed during my presentation last monday. It offers an additional option for computing the detection threshold when one has several epochs of data available, (or one wants to subdivide one epoch of data).
The idea is that for a larger set consisting of subsets of images, each single set could be described better by their own PC-bases. In this implementation the residuals can be computed for each set individually with their own PC-basis instead of a common PC-basis of a larger set of images.
This has been implemented by enabling multiple input tags, these tags can then be concatenated(`mode=combined`) (previously implemented) or used individually(`mode=individual`) (new implementation). When  `mode=individual` the residuals and noise are computed individually, before being passed to the `util.contrast_limit` function. This function has been adapted such that it can compute the flux and noise for the sets individually and combine them. Then the fake planet is injected for each set individually, the residuals are computed and the residuals are computed and combined. With all this information the contrast is then computed as before. Let me know if you have any further questions.

> The code has become quite long and complex, perhaps move some of the features to a separate util function? The `SimplexMinimizationModule` and `MCMCsamplingModule` may also need the features for the 'weighted' and 'clipped' residuals.


I will see how I can best refactor the `ContrastCurveModule`. I will look into these two modules. I will make a separate PR for those.

tomasstolker:
Thanks for the clarifications!

How is the PSF template handled in the new implementation?

> When mode=individual the residuals and noise are computed individually, before being passed to the util.contrast_limit function. This function has been adapted such that it can compute the flux and noise for the sets individually and combine them.

This part I don't fully understand... We can also discuss offline if that is easier.

benassa-de-glassa:
The psf template is handled same as before. Just the checking if the psf template comes from a input image has changed to make sure the same data is not connected to two input_ports.

> > When mode=individual the residuals and noise are computed individually, before being passed to the util.contrast_limit function. This function has been adapted such that it can compute the flux and noise for the sets individually and combine them.
> 
> This part I don't fully understand... We can also discuss offline if that is easier.

I'll swing by your office in the coming days.

tomasstolker:
For now, I don't think that we should add these additional features to the `ContrastCurveModule` since it will add quite some complexity which for the majority of the users is not required. Instead, I would propose that you leave these implementations on a branch of your fork such that those interested can use it from there. We could also make a note somewhere in the documentation that links to your fork. I will close this PR for now.

PR #377, opened 2019-06-24
weighted residuals for SimplexMinimization and MCMC

benassa-de-glassa:
As requested in #374 the `SimplexMinimizationModule` and the `MCMCsamplingModule` now support the `residuals=weighted` input.
Note: The `MCMCsampingModule` has the function `gaussian_variance` function commented, therefore the change has currently no effect. I don't know whether or not you want to uncomment this.


tomasstolker:
Great, thanks! The residuals in the `MCMCsamplingModule` are combined in `util.mcmc`. The commented code of `gaussian_variance` may be ignored.

benassa-de-glassa:
Okay, I'll have a look at `util.mcmc`. :D

PR #381, opened 2019-07-09
Reduction of memory usage by PcaPsfSubtraction module

aboehle:
In this pull request, I made a number of changes to the `PcaPsfSubtractionModule` to reduce the RAM usage (down to ~60% of the original memory usage for a testing data set of 6000 frames). 

First, I removed a few instances of unneeded copying or saving of data into separate variables. Specifically, the `images` input in the `pca_psf_subtraction` function is no longer copied and the PCA model in that same function is no longer saved in the separate variable `psf_model` prior to creating the image residuals. In `PcaPsfSubtractionModule` itself, a separate copy of the science data is no longer made in the case when the science data and the reference frames are the same input data tag.

I also changed `pca_psf_subtraction` so that only the de-rotated residuals are returned and not the still rotated residuals. The latter residuals were only used to calculate the variance for the weighted combination of the residuals images, so `pca_psf_subtraction` now outputs this variance instead of the full stack of residuals.

The `PcaPsfSubtractionModule`  also no longer removes the pixels with a value of 0 before performing PCA, to avoid making another copy of the science data when only pixels != 0 are selected. I found that for some cases (data sets of 60 -- 6000 frames for pca_numbers from 1 to 50) this change actually also improves the computation time, more so for the largest data sets (factor of 2 -- 3 for the testing data set of 6000 frames). Line profiling showed that the improvement is at the `np.matmul` step in `pca_psf_subtraction`, so perhaps this improvement has to do with the way the science data array (`im_reshape`) is contiguous in memory.


tomasstolker:
Thanks for the PR. The previous implementation which excluded masked pixels from the PCA calculation was done in order to speed up the computation time. Even though this makes a copy of the data, I think that including the masked indices in the PCA calculation will make it unnecessarily slow if for example half of the image is masked or if only an annulus is selected. I actually tested this when it was implemented and it made for me quite a difference in computation time. So I would suggest to change this back. Instead you could build in a check, for example if less than 90% of the data is masked then the full images are used for PCA.

aboehle:
Sure thing, I can add such a check to see if <90% of the pixels are masked before using the full frame for PCA. In my tests with a large number of frames I found that the PCA ran faster without removing the masked pixels, but in this case only a small fraction of pixels were masked.

tomasstolker:
Thanks for the improvements. I see now that there was a mistake in my previous message, I meant to say if less than 90% is unmasked instead of masked... I tried to change the threshold to an (arbitrary) < 80% (a more precise value could be determined in a separate PR if needed) but there appears an reshaping error in that case.

tomasstolker:
There appears a conflict with the master. Could you resolve the problem and sync the PR with the master? Thanks!

tomasstolker:
As mentioned offline, please reopen this PR from a new fork to sync with the updated repo history. Let me know if you have any questions. Thanks.

PR #382, opened 2019-07-17
Shift images to use multiprocessing

aboehle:
I've reverted the `ShiftImagesModule` back to using `apply_function_to_images`, now that the multiprocessing bug has been fixed. In doing speed/scaling tests, I found that shifting images in stacks is actually slower than looping over all the images and shifting them one by one, so the `ShiftImagesModule` is now faster both for the single and multiprocessing cases with this PR. I also sped up the way that the module checks if the x/y shift is constant for all frames.


tomasstolker:
Can you remove the `lastfailed` file and check the changed file with `pylint`? Then we can merge the PR.

PR #384, opened 2019-08-19
Introducing Iterative PCA Module

pahuber:
Implementation of reference-less algorithm of Pairet et al. (2018). Subtracts PSF of high-contrast ADI data of exoplanets and circumstellar disks minimizing the signal self-subtraction.


tomasstolker:
Hi @pahuber & @patapisp,

Thanks a lot for the PR. Great to have iPCA implemented in PynPoint. A few quick comments...

- Could you remove the *.npy and the PynPoint_config.ini files from the PR?
- Please have a look at the style of the autodocs and improve them accordingly.
- Could you also add type checks (see any of the other pipeline modules and util functions for examples)?
- The CI is no longer finishing. Please check what is wrong, all test cases should in principle finish. You can run them with pytest.
- Please also have a look at the CodeFactor issues that are easy to fix.
- If you have time, then it would also be useful to add another test case that used the iPCA option. You can probably copy-paste any of the PCA cases.

Let me know if you have any questions.

tomasstolker:
One more comment... I probably had misunderstood but I thought that iPCA added an additional parameter to the existing PCA module. Currently there is a whole new pipeline module for iPCA with mostly code duplication from the normal PCA pipeline module. Would it make sense to combine the two or will that become too complex?

patapisp:
@tomasstolker Thanks for the comments. I'll try to deal with them in the next 2 weeks.

tomasstolker:
Thanks for looking into this, let's try to finalize this PR when you are back. Let me know if you have any questions.

tomasstolker:
As mentioned offline, please reopen this PR from a new fork to sync with the updated repo history. Let me know if you have any questions.

Issue #386, opened 2019-08-26
Screen output from the ContrastCurveModule


tomasstolker:
The output printed to the screen has a small bug when the module has finished:

`Running ContrastCurveModule... [DONE]- ETA: 00:00:03`

which should be:

`Running ContrastCurveModule... [DONE]`

tomasstolker:
@benassa-de-glassa is this something you could fix? Thanks.

benassa-de-glassa:
yes, I'll have a look into it

benassa-de-glassa:
my proposed fix is in #387

tomasstolker:
I have fixed the problem with commit 6d82bd73724c42fce3b05327eca8da15227ba3db.

PR #387, opened 2019-08-26
time-display fix

benassa-de-glassa:
This is a pull request for issue #386 

The problem consists of two parts:

1. we are inconsistent throughout the package how we count measure the progress. Sometimes we start at 0, other times at 1. Sometimes we go to N, sometimes to N-1... depending on the configuration, we sometimes have the case where `current +1 > total` which was not covered by the if-clause in util.progress in line 73.. 
The new `>=` may lead to the time not beeing displayed for the very last iteration. We can revert this once we go through all loops where the `progress` function is called and standardize them.
I ran a couple of tests, the lack of the time display on the last iteration is only really noticable for very slow iterations.

2. As the contrast curve runs in parallel and we check every 5 seconds whether or not any progress has been made, it sometimes occurs that all the last tasks finish within the same 5 second window in which case we again go past the `current + 1 == total` if-clause.
I fixed this by using a callback function which gets called everytime a task finishes and increases the count of finished tasks. see lines 250-260 in `processing/limits.py`

### DIFF
--- a/pynpoint/processing/frameselection.py
+++ b/pynpoint/processing/frameselection.py
@@ -932,27 +932,29 @@ class FrameSimilarityModule(ProcessingModule):
         pool = mp.Pool(cpu)
         async_results = []
 
+        start_time = time.time()
+        nfinished = 0
+
+        # callback function accessing the nonlocal nfinished variable to keep track of the completed
+        # number of tasks
+        def plus_one(result):
+            # result gets passed into the callback even if we are not yet interested in it yet
+            # grab the non local counter which counts the number of computed postions
+            nonlocal nfinished
+            nfinished += 1
+            # print the progress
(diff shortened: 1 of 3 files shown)


benassa-de-glassa:
I don't understand why the coverage of `util/limits.py:contrast_limit` has decreased so much.. the tests in `tests/test_processing/test_limits.py` which use `util/limits.py:contrast_limit`  pass so I'm confused

tomasstolker:
I went through the pipeline modules to check the implementation of the `progress` function. It seems to me consistently implemented. The maximum value that is provided is always equal to the length of the for loop. The minimum value is always zero except for a few of the background modules so the printed output will start at an higher percentage in those cases.

tomasstolker:
This problem was fixed with commit 6d82bd73724c42fce3b05327eca8da15227ba3db so I will close this PR.

PR #400, opened 2019-12-18
PACO

nenasedk:
**Adding PACO analysis to pynpoint.**

Primary files:

- pynpoint/processing/pacoModule.py - _Wraps the PACO package into a pynpoint style module_
- pynpoint/util/paco.py - _Contains the PACO analysis, both the 'fast' and 'full' algorithms. Called by pacoModule and ContrastCurveModule_
- pynpoint/util/pacomath.py - _Some utility functions, mostly for computing covariance matrices, or positions._
- pynpoint/processing/limits.py - _Rewrote this module to have the option of using PACO to compute a contrast curve. If the 'name_in' argument contains 'paco', it will use PACO. I still do not fully trust the results from this._
- pynpoint/util/limits.py - _Effectively copied the contrast_curve function and modified it to use PACO._

Also updated are init files to import module names, and a couple test programs to ensure things run.

Everything should be compatible with the most up-to-date version of pynpoint. Merge conflicts will mostly come from a rewrite of the limits.py file. (Not exactly the most elegant solution for wrapping the two contrast curve modules, but it works)

**ISSUES:**
 - Contrast Curve using PACO needs to be tested more thoroughly
 - ROC curve module could be a future addition
 - PACO is still fairly slow, but that's mostly a limitation of the method.
 - Code is mostly documented, but likely some missing spots here or there
 - Need to write a 'How-to' so that it is useable to others.
 - Fairly sensitive to package versions, current requirements.txt works
 - tests/processing/test_PACO_FullPipeline.py shows required pre-processing steps. Should be more clearly documented what exactly pynpoint expects (ie number of rows/columns of images, centered etc)


tomasstolker:
Closing this duplicate PR for PACO since PR #401 is used for the actual implementation.

PR #401, opened 2019-12-18
PACO algorithm

nenasedk:
**Adding PACO analysis to pynpoint.**

Primary files:

- pynpoint/processing/pacoModule.py - _Wraps the PACO package into a pynpoint style module_

- pynpoint/util/paco.py - _Contains the PACO analysis, both the 'fast' and 'full' algorithms. Called by pacoModule and ContrastCurveModule_

- pynpoint/util/pacomath.py - _Some utility functions, mostly for computing covariance matrices, or positions._

- pynpoint/processing/limits.py - _Rewrote this module to have the option of using PACO to compute a contrast curve. If the 'name_in' argument contains 'paco', it will use PACO. I still do not fully trust the results from this._

- pynpoint/util/limits.py - _Effectively copied the contrast_curve function and modified it to use PACO.
Also updated are init files to import module names, and a couple test programs to ensure things run._

Everything should be compatible with the most up-to-date version of pynpoint. Merge conflicts will mostly come from a rewrite of the limits.py file. (Not exactly the most elegant solution for wrapping the two contrast curve modules, but it works)

**ISSUES:**
- Contrast Curve using PACO needs to be tested more thoroughly

- ROC curve module could be a future addition

- PACO is still fairly slow, but that's mostly a limitation of the method.

- Code is mostly documented, but likely some missing spots here or there

- Need to write a 'How-to' so that it is useable to others.

- Fairly sensitive to package versions, current requirements.txt works
tests/processing/test_PACO_FullPipeline.py shows required pre-processing steps. Should be more clearly documented what exactly pynpoint expects (ie number of rows/columns of images, centered etc)

- I'm _mostly_ sure I copied everything correctly from the previous fork.


tomasstolker:
Thanks a lot @nenasedk! This is really great :)!

A first few quick comments:

- Please run `pycodestyle` and `pylint` on the files that were added or changed and check for things that can be easily improved. Please also have a look at the `CodeFactor` issues in the PR.

- Please also run both `make docs` and `make test` in your local repo folder to make sure that both finish successfully. There seems to be a problem with the CI status in the PR.

- Please close the old/other PR on your earliest convenience.

nenasedk:
I'll take care of those. Thanks!

Cheers,
Evert

tomasstolker:
Hi @nenasedk, hereby a quick reminder about the open PR. Let me know if you have any questions about how to finish the implementation. There are two PRs at the moment, can you close the one that is not needed? Thanks :)

tomasstolker:
Hi @nenasedk, thanks a lot for all the improvements! I fixed the minor conflict with the documentation so the unit tests are now running on the Travis CI (but you can also run them locally as discussed offline). The status of the CI is shown at the bottom of this PR page. There are also some additional issues found by CodeFactor, could you have a look at the issues that are easy to solve?

nenasedk:
I'll take a look and get back to you.

tomasstolker:
Before we merge PACO, it would be great if (at least) one of the PynPoint users tries it out and provides some feedback in this PR. I currently don't have time to test the PACO module myself (but would like to in the future!). Is there anyone who is interested to take up this job?

18alex96:
I did not invest a lot of time for testing, but I encountered some problems when running the PACOmodule on my personal computer (MacBook Pro, 3.5 GHz Dual-Core Intel Core i7, 16 GB 2133 MHz LPDDR3 with macOS Caralina). I used more or less the default settings (fastpaco) on a SPHERE dataset of (66, 201, 201) frames with CPU=1 in the config file. For all test cases (slightly modifying the input parameters) the 'Process finished with exit code 137 (interrupted by signal 9: SIGKILL)' which is indicative of excessive memory usage, I suppose.
Are you aware of this issue and is it caused by some faulty setup from my side? Or is PACO intended only for use on machines with appropriate resources?

nenasedk:
I'll take a look. Right now I'd say the easiest fix would be to run it with
sufficient memory, but I'll see if there's a bug somewhere. There are a lot
of rather large matrices stored, so it's possible that your laptop doesn't
have enough memory for your test set. Maybe try cropping your images to
(66,101,101) or something, and see if it will at least run.

nenasedk:
@tomasstolker 
I changed some of the files that I had in the testing directory so that the pynpoint tests will actually run. I have NOT written unit tests to actually cover the PACO code, which is something that should probably be done. There were also a couple minor bugs I fixed when checking to make sure that everything runs properly.

@18alex96 
I don't have an image stack right now that's the same size as yours, but testing using 4 cores and 8GB of RAM lets me run a (61,101,101) stack with no issues. If you can, maybe try running on something where you can use ~32GB RAM, I figure by quadrupling the number of pixels you'd need about four times as much memory as well.  That said just watching my memory usage using htop, I never seemed to use more than maybe 1 or 2GB, so I'm not sure why you'd have issues. I'm using the. I've only ever tried running on linux though, so maybe there's a difference there somehow? I wouldn't expect that.

tomasstolker:
Thanks! Can you push the changes such that the PR is updated?

tomasstolker:
We should indeed add unit tests, but let's test first if the existing unit tests finish on Travis.

18alex96:
Sorry for my slow responsiveness. I was (and still am) a bit occupied with preparations for ESO proposals. I will have another look into this next week. But feel free to proceed without my feedback for now.

nenasedk:
Hey Tomas, I've pushed all of the necessary changes. The merge is blocked just by a few things in the codefactor, but I don't think there's anything very important in there. Since I haven't written unit tests obviously the test coverage has decreased as well. I've got my branch fully up to date with pynpoint, but it's been a bit of a confusing git mess trying to get everything sorted out there - I think I've sorted any issues though.

18alex96:
I finally had another chance to test the PACOModule and this time it worked without raising a memory error (not sure why this has happened before). A comparison between ADI+PCA (1 component, left) and the PACO SNR (right) is attached below.
Some preliminary suggestions for the implementation, based on my very limited testing so far:
- Something is weird with the position angles of the detected sources. Is this intended or are you aware of this behavior? Some sources are detected at the same radial separation, but somehow with a deviating position angle. This deviation is strongest for the source highlighted by the white arrow (I think it should be the same).
- `psf_rad` is required in arcsec if I am not mistaken. i would add this to the documentation. First, I thought pixel units are required. If `psf_rad` is required in arcsecond, I am wondering why the default value is 4'', which seems to be rather large? For my analysis I just chose a radius equivalent to the PSF FWHM of ~50 mas.
- `algorithm`: maybe add a quick description of the differences between fast and full PACO. Can you also add a rough estimate how the choice of algorithm affects the runtime?
- `flux_calc`: If you set `flux_calc` to `False`, an error is raised in line 169 of `pacoModule.py`, because `self.m_flux_out_port` is not defined. Perhaps, add check if the flux is calculated and needs to be written.
- If I am not mistaken, there is currently no opportunity to add a `psf_scaling` that corrects for different exposure times between science and flux images and potential ND filters? You might want to add some parameter like this to derive physical meaningful fluxes.
- output image size: is there any reason that the output image size is smaller than the input size? In my case the images were cropped from (201x201) to (200x200). If possible, I would suggest to keep the input size.
Hope this helps. I am happy to provide more feedback if I find some time to thoroughly test your module.
[paco_test.pdf](https://github.com/PynPoint/PynPoint/files/4552835/paco_test.pdf)

nenasedk:
Hey Alex,

I'll take a look at those issues and address some of them here.

Position angle: definitely possible that I made a mistake there.

Psf_rad: I'll try to improve the documentation and defaults. For a long
time I had it set to take the psf radius in pixels, so a radius of 4px was
set as the default.

Flux_calc: I forgot an if statement, I'll fix that.

Psf_scaling: I don't think that one will work with the way PACO estimates
flux right now. I actually normalize the I put psf, and paco finds a
scaling factor to find the flux. But you're right, I need to add a way to
ensure that I get physical units out.

Output image size: I had to do some funny stuff to the data to get it to
work with the sub pixel scaling that can be used - which meant enforcing
even dimensions. I'm sure there's a better way to do it, and that this
isn't necessary. The actual PACO people are much cleverer than I am, and
now have it set up to only do subpixel scaling in a small region around the
expected position of the companion(s). This would be a much better way of
doing things than the way it's currently implemented.

tomasstolker:
Thanks! A few more quick comments:

- Please also have a look at the issues from CodeFactor, these seem to be easy to fix.

- Could you check that all classes and functions have docstrings (including the parameters, returns, types, and descriptions), typehints, and typechecks?

- Why is it not possible to use odd-sized images? In case the images have odd dimensions, does the PACO module remove 1 line at apply a 0.5 pixel shift in both dimensions to recenter?

Let me know if you have any questions!

tomasstolker:
I have started to add unit tests for PACO. This is with a simple test data set for ADI (star  + offset planet). When running fastPACO, the SNR and flux output contain only NaNs. Perhaps this dataset is too simplistic for what PACO does? It would be good to understand what is going wrong in this case and add a warning with brief instruction for the user.

The small, test dataset runs without any memory problems but I also tested on a dataset with the shape (70, 113, 113) which got killed halfway on my macbook. Perhaps one could use a profiler to understand what the bottleneck of the module is to see if there is a way the solve this issue. Otherwise, it would be useful to add some instructions in the documentation about the ~maximum array size that can be processed.

nenasedk:
Hi @tomasstolker 

I've done some more work getting PACO running - I've made it fast enough to actually be useful, and fixed a few bugs that have been pointed out. I've gone through and fixed most of the CodeFactor issues, apart from where it seems to not like some of my docstrings for some reason. I'm also having trouble with the Travis integration tests - it's failing on the assertion that len(warning) == 3 after running both fast and full paco. Why is the length of the warnings supposed to be 3? It's saying the length is 2 after running fast paco, and 4 after full paco. Any ideas on how to fix this?

tomasstolker:
Hi @nenasedk,

I have changed the default branch of PynPoint from `master` to `main` but this closed your PR. Could you create a new PR on the `main` branch in case you or anyone else still still wants to finish it? Thanks and sorry for the inconvenience.

PR #402, opened 2020-02-11
ResidualSelectionModule, new parameters, maintenance

tomasstolker:
- Bug fix in `apply_function_to_images` with `MEMORY=None`.
- `offset` parameter in `SimplexMinimizationModule` and `SystematicsErrorModule`.
- `gaussian_noise` function for estimating the noise within an annulus after PSF subtraction and rotating the images in opposite direction.
- Additional printed output from `MCMCsamplingModule` and `SystematicsErrorModule`.
- Replaced `n_angles` with `angles` in `SystematicsErrorModule`.
- Some refactoring of `FrameSelectionModule`.
- `ResidualSelectionModule` for frame selection after the PSF subtraction.
- `angle_average`, `angle_difference`, and `stack_angles` functions.
- Updated documentation.
- Updated `requirements.txt`


tomasstolker:
Also added:

- `max_rotation` parameter in `StackAndSubsetModule` to set a maximum amount of field rotation per subset of stacked images
- unit tests for the new parameters and pipeline module

PR #410, opened 2020-05-07
Sdi

Kiefersv:
I added SDI support to the PcaPsfSubtractionModule. This includes changes in the pca_psf_subtraction function as well as the multiprocessing functions. Additionally, ADI and SDI were combined to create advanced differential imaging techniques.

The code style was checked and tests similar to the ones for ADI were added. All changes are backwards compatible.


tomasstolker:
Thanks a lot for the PR! Great work!

I didn't look at the implementation in detail yet but the code looks overall well written. For now, I have briefly tested the PCA module on a SPHERE/IFS dataset, so here a few comments:

- When using SDI+ADI or ADI+SDI I would suggest that the output of the residuals is always 5D (as discussed offline), also in case a single PC is used for either of the two steps. I tested with 1 PC for both, in which case the residuals reduced to 3D.

- Would it be easy to implement the ADI option also for IFS data? In that case ADI could be done on all wavelength channels separately.

- When using only ADI or SDI on an IFS dataset, I would suggest to make the output 4D, as is already the case for SDI.

- Some of the unit tests are broken. I think (but please check) that the issue is related to the output shape of the PCA residuals of non-IFS data. When using a single PC, the output should still be 3D.

- Running the PCA module with `CPU: 1` in the config file worked fine but using `CPU: 2` gave the following error: `TypeError: Can't broadcast (39, 151, 151) -> (151, 151)`.

I will have a more detailed look later on but I could use some help from any of the other PynPoint users! @18alex96, would you like to test the implementations and provide some feedback on the actual code? Anyone else interested to help?

18alex96:
Thanks a lot for your work on this, @Kiefersv and @tomasstolker ! I will have a look at the implementation in the coming days and I will try to provide feedback by the end of the week.

tomasstolker:
Perfect thanks!

A brief clarification to those who like to test PynPoint on IFS data: while @Kiefersv has already adjusted (locally) many of the PynPoint modules for compatibility with 4D datasets (i.e. including the wavelength dimension), we agreed to implement things stepwise.

As a first step, we assume that the data has already been preprocessed, and parallactic angles and calibrated wavelengths are available.

In PR #404, I had adjusted the `dataio` core of PynPoint for compatibility with IFS/4D data, as well as `FitsReadingModule`, which now has the (boolean) `ifs_data` parameter. I also added the `WavelengthReadingModule`, which can be used in tandem with `ParangReadingModule`.

With the current PR, the `PcaPsfSubtractionModule` has been adjusted by @Kiefersv such that an IFS dataset can be post-processed with SDI+ADI, ADI+SDI, SDI, or ADI.

Kiefersv:
Thanks for the kind words!

I implemented your comments and the output shape should now be fixed in all cases. I also checked the unit tests and the TypeError you mentioned but could no longer reproduce them. Additionally, I corrected ADI and it should now apply the correct processing for both IFS and single wavelength data.

18alex96:
I tested `ADI`, `SDI`, `ADI+SDI`, and `SDI+ADI` for a SPHERE/IFS dataset and the results look really nice, indeed. You can find the output that I obtained on a reasonably bright object [here](https://surfdrive.surf.nl/files/index.php/s/2059xAmIDiXjDZL) (Sven, you might know this one from your Master Thesis). If I am not mistaken, the `SDI` implementation seems to create some arbitrary shifts in x direction for some of the wavelength channels. For instance, when you have a look at `sdi_res_mean.fits` and slide though the wavelengths, the PSF seems to jitter along the horizontal axis (you can also notice this for the bad pixels). I did not figure out yet, what is causing this shift. Maybe this difference originates when scaling the images and the result is either even or odd sized? Let me know if you have some thoughts on this and if you have encountered this issues as well.

tomasstolker:
Thanks for working on the unit tests for the SDI module, @Kiefersv. I have made some further improvements and made the arrays with the test data a bit smaller. The runtime of the SDI tests is a few seconds now.

Kiefersv:
Thanks Tomas!
I could extend the test coverage to over 95% again. I don't think I can reasonablly increase it much more because most of the untested lines are error cases.
I also looked into the x shift problem mentioned by @18alex96. I found a small interpolation error but it only causes a shift of less than sqrt(2) pixels in x and y direction. I don't think that this the explenation. Fortunately, the shift shown by @18alex96 does not effect detections significantly.

tomasstolker:
Thanks for sharing the results, @18alex96.

There seems indeed an issue with the centering in the SDI residuals. This is something that should be fixed before merging the PR.

What is the negative feature in the SDI residuals that seems to move outward (across the inner point source) with wavelength? A bright speckle that isn't fully corrected? The morphology of the inner point source looks peculiar because of that, but the background objects seem fine.

tomasstolker:
@Kiefersv, perhaps you already did this, but can you test the PCA module both on odd- and even-sized images to make sure that the interpolation is done correctly in both cases? Please use `util.image.center_pixel` and `util.image.center_subpixel` getting the (x, y) position of the image center.

Kiefersv:
So, I took a while, but I found the cause of the jitter effect and corrected it. It was caused by the up and down scaling to either even or odd images during the SDI step. Unfortunately, the current correction also slightly reduces the quality of the SDI reduction.

Furthermore, I changed the scaling functions and they now use `util.image.center_subpixel` and `util.image.center_pixel`.

@18alex96: could you use the current version on your data and check if the jitter is gone?

tomasstolker:
Thanks for solving the issue with the scaling!

> Unfortunately, the current correction also slightly reduces the quality of the SDI reduction.
That seems counterintuitive. How did you determine that the quality is lower after the improved scaling?

Kiefersv:
That remark was a little misleading. The overall detectability of exoplanets using SDI should be improved (according to preliminarily tests). I corrected the jitter by adjusting the scaling factor to reduce interpolation errors caused by the two rescaling’s necessary for SDI. Therefore, I no longer use the perfect scaling factor for SDI, but the differences are very small.

tomasstolker:
Could you elaborate a bit? I don't understand what you mean with "I no longer use the perfect scaling factor for SDI".

Kiefersv:
Of course!

SDI uses two scaling factors. The first scaling factor S_1 is used to align the stellar PSF and create a model. This model is then subtracted from the raw (and equally scaled images). The second scaling factor S_2 is used to back scale all images and thus align a possible companion. The scaling factor for both steps are related: S_1 = 1 / S_2.

Now, the jittering is caused by interpolation inaccuracies. If the images are upscaled and then downscaled again, the centre can be shifted up to (1 / sqrt(2)) pixel. Thus, the resulting images can have a relative shift of up to sqrt(2) pixel. This shift cannot be corrected through proper alignment as the shifts can be of subpixel size.

The solution to the problem is to ensure that the shifts of all images after each scaling step are roughly pixel sized. This can be achieved by assuring that each scaled image always has an odd image size (even image sizes would also work but it must be the same for all images). Important to notice because the image size is the same for all input images, it does not matter if their size is even or odd (Furthermore, it also works for images which are not square).

I implemented this solution by rounding the scaling factor such that the resulting image is the closest odd sized approximation to the “real image size after rescaling”. This adjustment is repeated for the second scaling step as well. Therefore, the images changes size as follows:
Raw(odd or even)    --- scaling 1 ---->    PSF Model (odd)    --- scaling 2 ---->    Final images (odd)

The downside of this method is that the adjusted scaling factor now aligns the stellar PSF not as good as the original scaling factor as it is slightly offset. This offset is small (< sqrt(2) pixel) and in most cases even smaller than 1 pixel. Therefore, the gain in signal through the better alignment should be higher than the loss due to the imperfect modelling of the stellar PSF. My tests agreed with this result and showed that it solves the jittering problem.

I hope this clarifies what I did.

tomasstolker:
Ok, I think that I understand now. To summarize, the problem is that after the scaling some wavelengths have an odd-sized image while others are even-sized, correct? Therefore, the star is either at the center of a pixel or at the corner of 4 pixels?

> I implemented this solution by rounding the scaling factor such that the resulting image is the closest odd sized approximation to the “real image size after rescaling”. This adjustment is repeated for the second scaling step as well. Therefore, the images changes size as follows:
Raw(odd or even) --- scaling 1 ----> PSF Model (odd) --- scaling 2 ----> Final images (odd)

I think that we should use the exact scaling factors and not approximate values.

> The downside of this method is that the adjusted scaling factor now aligns the stellar PSF not as good as the original scaling factor as it is slightly offset. This offset is small (< sqrt(2) pixel) and in most cases even smaller than 1 pixel. Therefore, the gain in signal through the better alignment should be higher than the loss due to the imperfect modelling of the stellar PSF. My tests agreed with this result and showed that it solves the jittering problem.

Where does the sqrt(2) pixel offset come from?

Introducing an error of ~1 pixel is quite a lot if one wants to do accurate astrometry, so I think that we need to change this implementation.

An easy solution would be to simply shift the images that are even-sized by sqrt(0.5) pixel in order to align them with the odd-sized images? That would require an additional interpolation but we don't loose the subpixel accuracy in that case.

You could also have a look at the `SDIpreparationModule` (for dual-band SDI) to see how it is implemented there.

@gcugno, any comments/suggestions?

Kiefersv:
> Ok, I think that I understand now. To summarize, the problem is that after the scaling some wavelengths have an odd-sized image while others are even-sized, correct? Therefore, the star is either at the centre of a pixel or at the corner of 4 pixels?

Yes that’s basically the problem.

> Where does the sqrt(2) pixel offset come from?

If I increase the image size by one pixel, everything in the image gets stretched. This effect is very small in the centre, but at the corner of the image, it displaces signals by up to one pixel in x and y. Thus, the maximal offset is sqrt(2) pixel. In general, the offset is far smaller than sqrt(2) (depending on scaling factor corrections and separation form the centre) as this is only the upper bound. 

> An easy solution would be to simply shift the images that are even-sized by sqrt(0.5) pixel in order to align them with the odd-sized images?

I looked at the `SDIpreparationModule` and it could solve the problem. I will investigate it closer and correct the SDI scaling again.

Kiefersv:
You were right, the `SDIpreparationModule` was the key to the problem! Thanks for the help!

The jitter is now corrected, and the scaling factors are the correct ones for the SDI. Therefore, astrometry should be possible, and the quality should be enhanced as well.

@18alex96: again, it would be nice if you could test it again on your data and see if everything works for you as well. Thanks in advance!

Kiefersv:
I forgot to mention: I corrected the jitter in exactly the same way the `SDIpreparationModule` does. All other changes were deleted and are no longer necessary.

tomasstolker:
Excellent! I will try to have a look early next week!

18alex96:
I tested the latest version on the same dataset and indeed the results are much better than before (no more shifts in the residual images).
If you want to have a look yourself you can inspect both 'new' and 'old' reductions [here](https://surfdrive.surf.nl/files/index.php/s/2059xAmIDiXjDZL) (upload is still in progress, so maybe you have to wait for a few minutes until all files are in the directory).
So, from my side the PR is good to go.
Great work @Kiefersv !

tomasstolker:
Thanks for testing, @18alex96. The ADI residuals look fine to me but in the SDI (or combination with ADI) residuals, the morphology of the inner point source seems still odd? Is that an issue with PynPoint or the data?

image

18alex96:
@Kiefersv Please correct me if I am mistaken here:
I think these residuals are due to the fact that we have a rather bright companion in this dataset. The PSF displacement for SDI is much less (in terms of FWHMs) than for a regular ADI dataset (maximum scaling is around 1.6 when IFS is operated in extended mode, I think). This leads to these oversubtractions as shown in the attached image. When collapsing the final images along the wavelength dimension, this kind of averages out (see files with `_lam_median` suffix in the same directory). I suppose that this effect is less severe for fainter companions (perhaps I can verify later this afternoon).

tomasstolker:
Right, that makes sense! The mean is subtracted internally in the PCA function of `sklearn` after the images are scaled and before creating the PCs. Naively, I would have expected a negative, arc-like feature in radial direction instead of a negative point source that moves outward with radius. But I don't have much intuition for SDI...

tomasstolker:
Actually, the back scaling has a wavelength dependent so it indeed makes sense!

tomasstolker:
Hi @Kiefersv,

Thanks again for your work on the SDI implementation. Overall the code looks fine, but I have refactored parts of it to reduce the complexity and remove redundant code. I also made some smaller improvements (e.g. removing unused function parameters, correcting/adding docstrings, etc.).

Here some additional feedback:

- I would suggest to change `_residuals` back to `combine_residuals` and move the `for` loop that is currently in `combine_residuals` to `PcaPsfSubtractionModule`.

- With `CPU=1` there is currently no progress bar. That would be helpful to have since it can be quite slow. Is this something that you could easily add? Perhaps there is some code to copy from any of the other modules.

- The unit tests only test a few of the processing types. Was that on purpose?

Could you check and test once more the code that was added in this PR, and have a look if you agree with the changes that I made? Make sure to pull in the changes first.

tomasstolker:
Here a test on a SPHERE/IFS dataset of HR 8799. These are the mean residuals of SDI+ADI for one of the H band channels. Planet e is not visible here but that is presumably because of the self-subtraction in the SDI/PCA part.

hr8799

Kiefersv:
Thanks for testing and refactoring!

Yes, the explanation of @18alex96 and your comments are both correct.

About your comments:

> I would suggest to change `_residuals `back to` combine_residuals` and move the for loop that is currently in `combine_residuals` to `PcaPsfSubtractionModule`.

I can do this, but that will add many if conditions to the `PcaPsfSubtractionModule` and the 'multipca' files. If this increase in complexity is preferable to the current solution, I will change it.

> With CPU=1 there is currently no progress bar. That would be helpful to have since it can be quite slow. Is this something that you could easily add? Perhaps there is some code to copy from any of the other modules

I will add it.

> The unit tests only test a few of the processing types. Was that on purpose?

Yes, this was done to optimize the run time of the test while still keeping the best possible coverage. If you look into the unit tests, you'll see that each processing type is used in either single pca or multi pca.

tomasstolker:
> I can do this, but that will add many if conditions to the PcaPsfSubtractionModule and the 'multipca' files. If this increase in complexity is preferable to the current solution, I will change it.

Are the if conditions that are already there not sufficient to separate the different cases? Otherwise maybe leave it like this for now.

> Yes, this was done to optimize the run time of the test while still keeping the best possible coverage. If you look into the unit tests, you'll see that each processing type is used in either single pca or multi pca.

The ADI+SDI case is missing. The PSF subtraction unit tests run within a few seconds so I would suggest to add all 4 processing types to both the single and multiprocessing case.

Kiefersv:
> Are the if conditions that are already there not sufficient to separate the different cases? Otherwise maybe leave it like this for now.

No, because there are three cases: SDI, ADI or (SDI+ADI or ADI+SDI) which all produce different outputs and require different handling. Alternatively, I could write the code more concise to reduce the complexity at the cost of making it harder to read. 

> The ADI+SDI case is missing. The PSF subtraction unit tests run within a few seconds so I would suggest to add all 4 processing types to both the single and multiprocessing case.

Yes, you deleted that one when you refactored the code. It was tested with the Wsap or Tsap keyword. I will add this test again with the ADI+SDI keyword.

> With CPU=1 there is currently no progress bar. That would be helpful to have since it can be quite slow. Is this something that you could easily add? Perhaps there is some code to copy from any of the other modules

I checked the code and there should already be a progress bar (see psfsubtraction.py line 348). Maybe you used it with only one given PC? In that case no progressbar will be shown (and there wasn't one in the old version either).

tomasstolker:
> No, because there are three cases: SDI, ADI or (SDI+ADI or ADI+SDI) which all produce different outputs and require different handling. Alternatively, I could write the code more concise to reduce the complexity at the cost of making it harder to read.

OK let's leave it as it as.

> Yes, you deleted that one when you refactored the code. It was tested with the Wsap or Tsap keyword. I will add this test again with the ADI+SDI keyword.

Let's stick with the keywords that we had agreed on (ADI, SDI, ADI+SDI, SDI+ADI) and also use these keywords in the unit tests. That will be least ambiguous.

> I checked the code and there should already be a progress bar (see psfsubtraction.py line 348). Maybe you used it with only one given PC? In that case no progressbar will be shown (and there wasn't one in the old version either).

Right, that is probably what I did.

Good, I think that the PR is ready after the unit tests have been updated.

Kiefersv:
Thats amazing to hear! I have added the necessary unit tests and it should be ready.

Kiefersv:
Found some small mistakes in the sdi unit tests. Now, everything should be ready.

pynpoint/util/postproc.py
```diff
@@ -0,0 +1,158 @@
...
+                                                               pca_number=pca_number[0],
+                                                               pca_sklearn=pca_sklearn,
+                                                               im_shape=im_shape,
+                                                               indices=indices)
+
+    elif processing_type == 'SDI+ADI':
+        # SDI
+        res_raw_int = np.zeros(res_raw.shape)
+
+        for i, _ in enumerate(images[0]):
```
tomasstolker, line 111: Change this to something easier to undertsnad? E.g. `for i in range(image.shape[1])`. Also in the other for loops below. Not sure if my suggestion is correct though.

pynpoint/util/multipca.py
```diff
@@ -145,20 +161,34 @@ def run_job(self,
...
+        residuals, res_rot = postprocessor(images=self.m_star_reshape,
+                                           angles=self.m_angles,
+                                           scales=self.m_scales,
+                                           pca_number=pca_number,
+                                           pca_sklearn=self.m_pca_model,
+                                           im_shape=self.m_im_shape,
+                                           indices=self.m_indices,
+                                           processing_type=self.m_processing_type)
+
+        if self.m_processing_type == 'ADI' and res_rot.ndim == 3:
```
tomasstolker, line 178: Add a comment to explain this if conditions

pynpoint/util/multipca.py
```diff
@@ -145,20 +161,34 @@ def run_job(self,
             Output residuals.
         """
 
-        residuals, res_rot = pca_psf_subtraction(images=self.m_star_reshape,
-                                                 angles=self.m_angles,
-                                                 pca_number=int(tmp_task.m_input_data),
-                                                 pca_sklearn=self.m_pca_model,
-                                                 im_shape=self.m_im_shape,
-                                                 indices=self.m_indices)
+        if isinstance(tmp_task.m_input_data, tuple):
```
tomasstolker, line 164: Add a comment to explain this if condition

pynpoint/processing/psfsubtraction.py
```diff
@@ -189,67 +276,193 @@ def _run_multi_processing(self,
...
+                        self.m_res_arr_out_ports[pca_1].add_history(
+                            'PcaPsfSubtractionModule', f'max PC number = {pca_first}')
 
-            # 1.) derotated residuals
-            if self.m_res_arr_out_ports is not None:
-                self.m_res_arr_out_ports[pca_number].set_all(res_rot)
-                self.m_res_arr_out_ports[pca_number].copy_attributes(self.m_star_in_port)
-                self.m_res_arr_out_ports[pca_number].add_history('PcaPsfSubtractionModule', hist)
+                    else:
+                        if len(pca_first)+len(pca_secon) == 2:
```
tomasstolker, line 369: Remove this if condition. Instead, add an ValueError in the class `__init__` in case of IFS data and there are more than 1 PC used in combination with res_arr_out.

pynpoint/processing/psfsubtraction.py
```diff
@@ -43,44 +45,76 @@ def __init__(self,
...
+            Database entry where the the noise-weighted residuals will be stored (see Bottom et al.
+            2017). The residuals are not calculated and stored if set to None.
         res_rot_mean_clip_tag : str, None
             Tag of the database entry of the clipped mean residuals. Not calculated if set to
             None.
         res_arr_out_tag : str, None
-            Tag of the database entry with the derotated image residuals from the PSF subtraction.
-            The tag name of `res_arr_out_tag` is appended with the number of principal components
-            that was used. Not calculated if set to None. Not supported with multiprocessing.
+            Database entry where the derotated, but not collapsed, residuals are stored. The number
```
tomasstolker, line 81: Explain how this works in case of IFS data.

pynpoint/processing/psfsubtraction.py
```diff
@@ -189,67 +276,193 @@ def _run_multi_processing(self,
...
+        # Configurate data output according to the processing type
+        # 1.) derotated residuals
+        if self.m_res_arr_out_ports is not None and len(pca_first)+len(pca_secon) == 2 and self.m_processing_type != 'Oadi':
+            if pca_secon[0] == -1:
+                history = f'max PC number = {pca_first}'
+
+            else:
+                history = f'max PC number = {pca_first} / {pca_secon}'
+
+            squeezed = np.squeeze(out_array_res)
```
tomasstolker, line 435: What is happening here? Please add a comment.

pynpoint/processing/psfsubtraction.py
```diff
@@ -189,67 +276,193 @@ def _run_multi_processing(self,
...
+                                                              res_rot=res_rot,
+                                                              angles=parang)
 
-            # 5.) clipped mean residuals
-            if self.m_res_rot_mean_clip_out_port is not None:
-                stack = combine_residuals(method='clipped', res_rot=res_rot)
-                self.m_res_rot_mean_clip_out_port.append(stack, data_dim=3)
+        # Configurate data output according to the processing type
+        # 1.) derotated residuals
+        if self.m_res_arr_out_ports is not None and len(pca_first)+len(pca_secon) == 2 and self.m_processing_type != 'Oadi':
```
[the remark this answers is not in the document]
Kiefersv: Yes. And since you added the self.m_ifs_data parameter, I could use that and also simplify line 362 in the same way.

pynpoint/processing/psfsubtraction.py
Kiefersv: I removed it and all other similar if conditions. I also added the requested ValueError.

pynpoint/processing/psfsubtraction.py
Kiefersv: Added the comment. Please check if it is understandable (as soon as I uploaded the changes).

pynpoint/processing/psfsubtraction.py
Kiefersv: I added the squeezing to ensure uniform data output for the residuals. We discussed earlier that ADI+SDI and SDI+ADI should always produce 5 dimensional outputs. If you want the same for residuals I can add it but since the first two dimensions are always 1 this might not makes sense.

For now I added comments explaining this.

pynpoint/util/multipca.py
```diff
@@ -283,12 +318,14 @@ def __init__(self,
                  weighted_out_port: Optional[OutputPort],
                  clip_out_port: Optional[OutputPort],
                  num_proc: int,
-                 pca_numbers: np.ndarray,
-                 pca_model: PCA,
+                 pca_numbers: Union[tuple, np.ndarray],
+                 pca_model: Union[PCA, None],
```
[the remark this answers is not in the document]
Kiefersv: Changed this one and all other `Union["something", None]` lines in the multipca.py file.

pynpoint/util/multipca.py
Kiefersv: Added comment and simplified the expression.

pynpoint/util/multipca.py
Kiefersv: Added the comment.

pynpoint/util/postproc.py
Kiefersv: I changed it in the suggested way.

pynpoint/processing/psfsubtraction.py
tomasstolker: Right, the current implementation seems fine, thanks.

PR #412, opened 2020-06-26
Psfpreparation adjustments for sdi

Kiefersv:
Now that the `PSFsubtractionModule` supports IFS data, I added IFS support to the modules of the psfpreparation.py file. The following changes were made:

- `PSFpreparationModule`: Added support for 4 dimensional IFS data. Added unit tests for IFS data.

- `SortParangModule`: Added support for 4 dimensional IFS data. Added unit tests for both IFS and non-IFS data (later was missing in the unit tests).

- `AngleCalculationModule`: Added support for 4 dimensional IFS data. Added unit tests for IFS data. Because PynPoint cannot handle raw SPHERE/IFS data, a preprocessing argument was added. This allows to read in data prepared with Pypelines such as EsoReflex. The EsoReflex angle calculation requires an additional attribute which is read in from the header: DATE_CORRECTION.


tomasstolker:
Thanks! I think that something went wrong with the branch that you created because the commits from the previous commits are there as well. Please close the current PR, update your master branch, and then create the new branch from the master branch.

Preferably with changes to one pipeline module at a time because in the "Files changed" it seems that a lot of the code has changed. Otherwise, it will be too much work for me to review the PR. Also, I think that it may be useful to first agree on the 4D update of a single module before implementing the other modules.

PR #413, opened 2020-06-26
SDI support for PSFpreparationModule

Kiefersv:
Added 4D IFS data support for the PSFpreparationModule including unit tests.

### DIFF
--- a/pynpoint/processing/psfpreparation.py
+++ b/pynpoint/processing/psfpreparation.py
@@ -99,21 +99,18 @@ class PSFpreparationModule(ProcessingModule):
         # Get PIXSCALE and MEMORY attributes
         pixscale = self.m_image_in_port.get_attribute('PIXSCALE')
         memory = self._m_config_port.get_attribute('MEMORY')
-
-        # Get the number of images and split into batches to comply with memory constraints
+        ndim = self.m_image_in_port.get_ndim()
         im_shape = self.m_image_in_port.get_shape()
-        nimages = im_shape[-3]
-        frames = memory_frames(memory, nimages)
 
         # Read out number of wavelength
-        if len(im_shape) < 4:
-            wavelength_dim = 1
+        if ndim < 4:
+            nimages = im_shape[-3]
+            frames = memory_frames(memory, nimages)
         else:
-            wavelength_dim = im_shape[-4]
+            time_chanels = im_shape[-3]
+            frames = np.linspace(0, time_chanels, time_chanels+1)
 
-            # Set output shape
-            tmp_output = np.zeros(im_shape)
-            self.m_image_out_port.set_all(tmp_output, keep_attributes=False)
+        # Get the number of images and split into batches to comply with memory constraints 
 
         # Convert m_cent_size and m_edge_size from arcseconds to pixels
         if self.m_cent_size is not None:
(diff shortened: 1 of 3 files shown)


pynpoint/processing/psfpreparation.py
```diff
@@ -104,9 +102,19 @@ def run(self) -> None:
 
         # Get the number of images and split into batches to comply with memory constraints
         im_shape = self.m_image_in_port.get_shape()
-        nimages = im_shape[0]
+        nimages = im_shape[-3]
```
tomasstolker, line 105: Add something like `ndim = self.m_image_in_port.get_ndim()` here and use the value of `ndim` for the `if` conditions where the number of dimensions (3 or 4) is required.

pynpoint/processing/psfpreparation.py
```diff
@@ -104,9 +102,19 @@ def run(self) -> None:
...
 
+        # Read out number of wavelength
+        if len(im_shape) < 4:
+            wavelength_dim = 1
+        else:
+            wavelength_dim = im_shape[-4]
+
+            # Set output shape
+            tmp_output = np.zeros(im_shape)
+            self.m_image_out_port.set_all(tmp_output, keep_attributes=False)
```
tomasstolker, line 116: Please do not use `set_all` or `get_all` unless strictly necessary. Instead, append the data to the output port in amounts of `MEMORY`.

pynpoint/processing/psfpreparation.py
```diff
@@ -123,32 +131,40 @@ def run(self) -> None:
...
-            images[:, ~mask] = 0.
+                # Apply the mask, i.e., set all pixels to 0 where the mask is False
+                images[:, ~mask] = 0.
 
-            # If desired, normalize the images using the Frobenius norm
-            if self.m_norm:
-                im_norm = np.linalg.norm(images, ord='fro', axis=(1, 2))
-                images /= im_norm[:, np.newaxis, np.newaxis]
-                norms.append(im_norm)
+                # If desired, normalize the images using the Frobenius norm
```
tomasstolker, line 157: Is this still correct with 4D data? Perhaps we should make the normalization (which I am anyway not a fan of...) only compatible with 3D data?

pynpoint/processing/psfpreparation.py
```diff
@@ -123,32 +131,40 @@ def run(self) -> None:
...
+                    images /= im_norm[:, np.newaxis, np.newaxis]
+                    norms.append(im_norm)
 
-            # Write processed images to output port
-            self.m_image_out_port.append(images, data_dim=3)
+                # Write processed images to output port
+                if len(im_shape) < 4:
+                    self.m_image_out_port.append(images, data_dim=3)
+                else:
+                    self.m_image_out_port[j, frames[i]:frames[i+1], ] = images
```
tomasstolker, line 167: If possible, then better to use `append` with `data_dim=4`, but please check if it is appended to the right axis. We can also update `core.dataio` if needed.

pynpoint/util/tests.py
```diff
@@ -299,6 +299,58 @@ def create_star_data(path: str,
         hdu.writeto(os.path.join(path, f'images_{j}.fits'))
 
 
+@typechecked
```
tomasstolker, line 302: Didn't we already add a function for test IFS data in the previous PR?

pynpoint/processing/psfpreparation.py
```diff
@@ -123,32 +131,40 @@ def run(self) -> None:
...
+                    warnings.warn('The input data has 2 dimensions whereas 3 dimensions are required. '
+                                  'An extra dimension has been added.')
 
-                images = images[np.newaxis, ...]
+                    images = images[np.newaxis, ...]
 
-            # Apply the mask, i.e., set all pixels to 0 where the mask is False
-            images[:, ~mask] = 0.
+                # Apply the mask, i.e., set all pixels to 0 where the mask is False
+                images[:, ~mask] = 0.
```
tomasstolker, line 155: In this case, the same mask is applied to all wavelengths? Should the size not increase linearly with wavelengths such that the masks overlap after the scaling in the PCA module?

pynpoint/processing/psfpreparation.py
Kiefersv: I can do this, but the -3 basically does this in one step instead of two if conditions.

pynpoint/processing/psfpreparation.py
Kiefersv: It applies the nomralization to each wavelength seperatly, so it should be similar to the 3D case. But I agree that only allowing it for 3D data is the best solution.

pynpoint/processing/psfpreparation.py
tomasstolker: Sorry I meant adding an extra line with `ndim =` and using the `ndim` value whenever the number of dimensions is required.

pynpoint/processing/psfpreparation.py
tomasstolker: OK let's add a warning in case the normalization is set to True with 4D data, and turn off the normalization in that case.

pynpoint/processing/psfpreparation.py
Kiefersv: The problem here is that the `memory_frames` only splits along 1 dimension (time) where IFS data has two dimensions (wavelength, time). I could work around this, but it might be cleaner if the `memory_frames` function would be adjusted.

The second problem is that I'm not sure if append works with IFS data. For non-IFS data, append can simply add an arbitrary number of frames along the time dimension. With IFS data, even though you can add along the wavelength dimension, you would need all time frames already prepared to be added. The only work around possible I found was the one I used in lines 164-167.

pynpoint/util/tests.py
Kiefersv: Yes we added a IFS data function for the PSFsubtractionModule unit tests. But since non-IFS test data was also generated with two functions, I did it accordingly.

pynpoint/processing/psfpreparation.py
Kiefersv: That makes sens. I'll change it.

pynpoint/processing/psfpreparation.py
Kiefersv: I added the warning and turned off normalization for 4D data.

pynpoint/processing/psfpreparation.py
Kiefersv: In the case of a non-scaling mask, the inner and outer radii are limited by the scaling of the PSF. If you cut to close around a possible planet, the PSF model will no longer be accurate. In the case of a scaling mask, the inner and outer radii are limited by the "radial displacements" of a possible exoplanet. If you cut to close around a possible planet, you'll mask out the planet in some wavelengths.

Since the scaling is the same in both cases, the area which should not be masked out is the same in both cases. Additionally, masking a stationary image instead of a scaling one is more intuitive and therefore easier to handle. I would therefore suggest leaving it as it is.

pynpoint/processing/psfpreparation.py
tomasstolker: Perhaps we could process and append the data per wavelength. So in case of IFS data, all time steps are processed at once. So it will not use `MEMORY`, but that is perhaps also less critical with NIR data.

pynpoint/util/tests.py
tomasstolker: Let's try to use the function that was already there.

pynpoint/processing/psfpreparation.py
tomasstolker: OK

pynpoint/processing/psfpreparation.py
Kiefersv: That is defentily a good option. I will implement it like this for now, but I'll keep thinking about it.

Kiefersv: I tried the implementation but using `self.m_image_out_port.append(images, data_dim=4)` appends along the second dimension instead of the first. Could you look into this?

pynpoint/processing/psfpreparation.py
Kiefersv: This problem is a part of the one from line 115. I'll solve this together with the other one and considre this instence as resolved.

pynpoint/util/tests.py
Kiefersv: I'll removed `create_star_data_ifs` and used `create_ifs_data` instead.

pynpoint/processing/psfpreparation.py
tomasstolker: Right, that is indeed the case in `dataio`. Maybe that is required by the PCA module, but I am not sure. We could add a parameter, but perhaps the easiest solution would be to process al the wavelengths per exposure at once (instead of the other way around, as I previously suggested)? Would that work?

Kiefersv: You are right! I miss understood you the first time but now implemented it as you mentioned it and it works fine.

tomasstolker:
I have made several minor improvements to the `PSFpreparationModule`.

Since `ndim` is always 3 or 4, I think that we can use these actual values in any if statements that test the value of `ndim` (instead of e.g. `ndim < 4`), to increase the readability of the code.

@Kiefersv, let me know if you have any further comments. Otherwise, we can merge the PR?

Kiefersv:
Thanks, the improvements look good. I have no further comments and it should be ready to merge.

PR #415, opened 2020-07-06
SDI support for SortParangModule including testing

Kiefersv:
I added SDI support for the SortParangModule and added unit tests for both IFS and non-IFS data.

### DIFF
--- a/pynpoint/processing/psfpreparation.py
+++ b/pynpoint/processing/psfpreparation.py
@@ -337,8 +337,9 @@ class SortParangModule(ProcessingModule):
 
         memory = self._m_config_port.get_attribute('MEMORY')
         index = self.m_image_in_port.get_attribute('INDEX')
+
         ndim = self.m_image_in_port.get_ndim()
-        im_shape = self.m_image_in_port.get_shape()
+        nimages = self.m_image_in_port.get_shape()[-3]
 
         index_new = np.zeros(index.shape, dtype=np.int)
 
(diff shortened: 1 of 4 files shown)


pynpoint/processing/psfpreparation.py
```diff
@@ -356,8 +359,6 @@ def run(self) -> None:
 
         index_sort = np.argsort(index)
 
-        nimages = self.m_image_in_port.get_shape()[0]
-
         frames = memory_frames(memory, nimages)
```
tomasstolker, line 362: Should this be adjusted for the 4D case, similar to the `PSFpreparationModule`?

Kiefersv: You are right. But this actually gives my an idea on how to optimize the frames problem with 4D data. I will investigated it and, if it works, upload it at a later time.

For now I changed the frames handling equivalent to the one used in the PsfPreparationModule.

tomasstolker: OK, thanks. The `i` index is already used within the for loop in which it is embedded (should have been picked up by `pylint`?). The `enumerate` can probably removed (see my previous commit).

Kiefersv: You are right. I left it because it was previously done this way. I now simplified the expression and corrected a minor bug.

tomasstolker:
I made a few improvements in the documentation and have indicated the pipeline modules that support IFS data on the following page:

https://pynpoint.readthedocs.io/en/latest/overview.html

PR #416, opened 2020-07-07
SDI support for the AngleCalculationModule

Kiefersv:
I added IFS support (4D data) to the AngleCalculationModule including testing. The following changes were necessary and might need to be looked at into more detail:
- I added the argument `preprocessing` to allow the angle calculation of data preprocessed with the EsoReflex piepline.
- I added the attribute `DATE_CORRECTION` which is necessary for correct date asignment if the data set is preprocessed with EsoReflex.


Kiefersv:
**Just a quick note:** I know that 3 of the tests currently fail. I will adjust them as soon as we agree on how the implementation should look.

tomasstolker:
Thanks for the PR!

I am not sure what is the best approach here. In general, I am a bit reluctant to optimize the `AngleCalculationModule` to the output of `esoreflex` (does anyone use that for SPHERE data anyway?).

What does the output of `esoreflex` look like? Are the original FITS headers still available in the preprocessed data?

Kiefersv:
I understand your concerns, but I nevertheless think it is necessary. PynPoint is currently not able to split raw IFS data into its wavelength components therefore this step has to be done before using PynPoint. We also discussed alternatives to EsoReflex (e.g. VLTPF) and I would recommend including them as well.

The output of EsoReflex is similar but (at least in my case) the `DATE-OBS` is assigned "incorrectly". The input is of the form (FRAMES, X, Y) and the outputs are FRAMES files of the form (LAMBDA, X, Y) which all have the timestamp of the first frame. Using the `DATE_CORRECTION` parameter the `DATE` can be assigned correctly.

tomasstolker:
> I understand your concerns, but I nevertheless think it is necessary. PynPoint is currently not able to split raw IFS data into its wavelength components therefore this step has to be done before using PynPoint. We also discussed alternatives to EsoReflex (e.g. VLTPF) and I would recommend including them as well.

Yes, I would recommend using VLTPF instead of EsoReflex. In that case, the angles are stored as FITS file, which can be given as input to `ParangReadingModule`.

> The output of EsoReflex is similar but (at least in my case) the DATE-OBS is assigned "incorrectly". The input is of the form (FRAMES, X, Y) and the outputs are FRAMES files of the form (LAMBDA, X, Y) which all have the timestamp of the first frame. Using the DATE_CORRECTION parameter the DATE can be assigned correctly.

How does the `DATE_CORRECTION` parameter work? Does it also correct for having multiple integrations per (raw) FITS file?

There might be more issues with EsoReflex. For example, I understood that the wavelength calibration is not accurate.

Kiefersv:
> Yes, I would recommend using VLTPF instead of EsoReflex. In that case, the angles are stored as FITS file, which can be given as input to ParangReadingModule.

And how are the Wavelengths stored? And is the `WavelengthReadingModule` already compatible with the VLTPF prepared data? 

> How does the DATE_CORRECTION parameter work? Does it also correct for having multiple integrations per (raw) FITS file?

Yes, that’s what I meant with FRAMES as I couldn't remember the correct term. The `DATE_CORRECTION` includes the information which integration number the image belongs to and the AngleCalculationModule corrects accordingly.

> There might be more issues with EsoReflex. For example, I understood that the wavelength calibration is not accurate.

Ok. Therefore, EsoReflex might not be suitable for precise measurements. Nethertheless, it should still be good enough for exoplanet detections using SDI+ADI (or ADI+SDI).

Tell me if you want to include EsoReflex or not and I will adjust the code accordingly (in both cases there are adjustments necessary). In any case, I would highly suggest that we check/add VLTPF support.

tomasstolker:
> And how are the Wavelengths stored? And is the WavelengthReadingModule already compatible with the VLTPF prepared data?

The wavelengths from VLTPF are also stored as FITS file and these can be read by the `WavelengthReadingModule`.

> Tell me if you want to include EsoReflex or not and I will adjust the code accordingly (in both cases there are adjustments necessary). In any case, I would highly suggest that we check/add VLTPF support.

Angles and wavelengths can be imported as FITS or ASCII files, so it does already support the output from VLTPF. That pipeline seems to me superior to EsoReflex, so I think that at this point we do not need to support the output from EsoReflex, but I never used EsoReflex myself so any feedback from other users would be welcome.

Kiefersv:
> The wavelengths from VLTPF are also stored as FITS file and these can be read by the WavelengthReadingModule.

Nice!

> Angles and wavelengths can be imported as FITS or ASCII files, so it does already support the output from VLTPF. That pipeline seems to me superior to EsoReflex, so I think that at this point we do not need to support the output from EsoReflex, but I never used EsoReflex myself so any feedback from other users would be welcome.

Ok. So I leave the AngleCalculationModule for the moment and will adjust it if you get postive feedback regarding the EsoReflex pipeline. If that's all, I will close this pull request.

tomasstolker:
OK agreed, let's keep it in mind, but I think it's fine to not implement it for now.

PR #418, opened 2020-07-08
SDI support for WaffleCenteringModule

Kiefersv:
I add SDI support for the WaffleCentringModule. No unit tests are added yet as only very few lines differ between IFS and non-IFS data. Furthermore, I add a new parameter `angle` which allows to use all waffle patterns (not only `x` or `+`). Please take a closer look at line 1019. I was not sure how to store the center information correctly as it can differ between wavelengths.


tomasstolker:
Thanks for this new PR, @Kiefersv!

@18alex96, would you mind reviewing this PR since you wrote the `WaffleCentringModule`?

Why is the `angle` parameter required? I did not know that it was possible to rotate the satellite spots in an arbitrary orientation, but my experience might be too limited.

How was the accuracy of the centering tested for IFS data? It might be useful to add unit tests because with test data it is easier to identify an issue with the accuracy of the subpixel centering.

Kiefersv:
> Why is the angle parameter required? I did not know that it was possible to rotate the satellite spots in an arbitrary orientation, but my experience might be too limited.

The FoV of IFS data is rotated 55 degrees.

> How was the accuracy of the centering tested for IFS data? It might be useful to add unit tests because with test data it is easier to identify an issue with the accuracy of the subpixel centering.

I will add tests but this might require adding a new function to the test.py file. Would that be ok? Furhtermore, I only changed the data handling but none of the actual centring process. Should I add testing only for the handling or both?

tomasstolker:
> The FoV of IFS data is rotated 55 degrees.

Then I would suggest to only use the `angle` parameter and deprecate the `pattern` parameter in a future release. I would set the default value of `pattern` to None and add a `DeprecationWarning` if the used value is not `None` to inform the user about these changes. Then please make sure to update the docstrings and explain that the '+' pattern is used with `angle=0', the 'x' pattern with `angle=45`, and for SPHERE/IFS `angle=55` is required. I would set the default value to 45. @18alex96, would you agree with this?

> I will add tests but this might require adding a new function to the test.py file. Would that be ok? Furhtermore, I only changed the data handling but none of the actual centring process. Should I add testing only for the handling or both?

OK let's leave it as it is for now, the unit tests for regular imaging should be sufficient then.

Kiefersv:
> Then I would suggest to only use the angle parameter and deprecate the pattern parameter in a future release. I would set the default value of pattern to None and add a DeprecationWarning if the used value is not None to inform the user about these changes. Then please make sure to update the docstrings and explain that the '+' pattern is used with angle=0, the 'x' pattern with angle=45, and for SPHERE/IFS angle=55 is required. I would set the default value to 45. @18alex96, would you agree with this?

I changed it as you suggested.

18alex96:
I had some time to test the centering module and the results look nice to me. Some feedback:
- `radius`: Add a quick description that for IFS data the minimum separation (aka shortest wavelength) is required. Apparently this is exactly the default value of 45 pixels. But it was not directly clear to me what to enter here without checking the code.
-  `pattern` versus `angle`: My personal preference would be to keep the pattern keyword, because that is also what you enter in P1 when designing your OBs. The IFS just has an additional offset of 100.48 degrees compared to IRDIS (see SPHERE manual, not sure where the 55 come from?). So it would be easier just to add this value in case of IFS data (`ndim == 4`). With the current implementation I was able to hit the reference spots exactly with `pattern=None` and `angle=45+100.48` (for an 'x' waffle pattern dataset according to P1). But I do not have strong feelings about this, if you prefer the `angle` parameter instead. Having an angle parameter comes along with more freedom (for instance when you want to reduce GPI data; which I never did, so I have no idea how the waffle spots are oriented there).

Other than this I do not have any comments and the images look nicely centered for all wavelengths and time instances. Great job @Kiefersv

Kiefersv:
Thank you very much @18alex96 for looking through the changes! 

> radius: Add a quick description that for IFS data the minimum separation (aka shortest wavelength) is required. Apparently this is exactly the default value of 45 pixels. But it was not directly clear to me what to enter here without checking the code.

I will add a description.

> pattern versus angle: My personal preference would be to keep the pattern keyword, because that is also what you enter in P1 when designing your OBs. The IFS just has an additional offset of 100.48 degrees compared to IRDIS  ...

I agree that it depends on the usage of the pipeline. Simply setting the angle to SPHERE/IFS data if the input data is 4D could lead to future mistakes if other 4D data is read in. I would suggest leaving either the angle parameter or to add a "SPHERE/IFS" pattern symbol. @tomasstolker: What would you prefer?

tomasstolker:
To me it seems fine to make the module more general and not tune it to SPHERE. I am also not sure how the module could recognize SPHERE/IFS data, otherwise? Similar with the PSF subtraction, we have the `extra_rot` parameter for a TN correction, which is not tuned to a specific instrument.

> The IFS just has an additional offset of 100.48 degrees compared to IRDIS (see SPHERE manual, not sure where the 55 come from?)

It probably depends if the rotation is done relative to the plus or cross pattern, because it is a difference of 45 deg.

> I would suggest leaving either the angle parameter or to add a "SPHERE/IFS" pattern symbol. @tomasstolker: What would you prefer?

I like the angle parameter, as long as it it clearly explained in the docstrings.

18alex96:
No objections from my side. Just go ahead with the angle parameter, which makes the module more flexible.

Kiefersv:
I added the requested comments and the WaffleCenteringModule should now be ready.

PR #419, opened 2020-07-28
SDI support for CropImagesModule

Kiefersv:
I added SDI support to the CropImagesModule. The changes are equivalent to the ones used in the PSFpreparationModule. I added no testing as only the data handling was changed.


tomasstolker:
Thanks for this new PR! Looks good to me, but let's add a unit test for the modules that are updated for compatibility with 4D data. That should hopefully be straightforward when using the `create_ifs_data` function and adopting an already existing unit test for `CropImagesModule`.

Kiefersv:
So, after a week of batteling with the sys admins of the cluster I'm working on, pytest-3.7 does somewhat work again. 

I added unit tests for IFS data handling.

tomasstolker:
Thanks for adding the unit test, @Kiefersv!

PR #421, opened 2020-08-11
SDI support for DerotateAndStackModule

Kiefersv:
I added IFS (4D data) support to the DerotateAndStackModule incl. unit tests.

What needs to be discussed is the output format of 4D data after stacking. Previously, the Module simply collapsed the time axis
 - e.g. (10, 291, 291) -> (1, 291, 291).

For 4D data, two outputs are possible:
- Case 1: (39, 10, 291, 291) -> (39, 1, 291, 291)
- Case 2: (39, 10, 291, 291) -> (39, 291, 291)

Case 2 might be less intuitve but has the advantage that the DerotateAndStackModule can now also be used to stack the data along the wavelength dimension. If the output form should be case 1 then we should add a parameter to the DerotateAndStackModule to stack along the wavelegnth axis.

Please tell me which one you prefere (currently case 2 is implemented).

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -129,7 +129,7 @@ Frame Selection
 Image Resizing
 ~~~~~~~~~~~~~~
 
-* :class:`~pynpoint.processing.resizing.CropImagesModule`: Crop the images.
+* :class:`~pynpoint.processing.resizing.CropImagesModule` (IFS): Crop the images.
 * :class:`~pynpoint.processing.resizing.ScaleImagesModule` (CPU): Resample the images (spatially and/or in flux).
 * :class:`~pynpoint.processing.resizing.AddLinesModule`: Add pixel lines on the sides of the images.
 * :class:`~pynpoint.processing.resizing.RemoveLinesModule`: Remove pixel lines from the sides of the images.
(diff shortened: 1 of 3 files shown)


tomasstolker:
Great, thanks @Kiefersv!

Since we now explicitly check in the updated modules if the input data is 3D or 4D, I am somewhat favoring case 1. If the data is passed to another pipeline module after running `DerotateAndStackModule` on an IFS dataset, then the shape of the data is still correctly interpreted (even though there is perhaps no further processing to be done after stacking the time dimension).

Perhaps we can indeed add a parameter (e.g. `dimension`?) which can be set to `'time'` or `'wavelength`'? Then both dimensions can be stacked separately.

Kiefersv:
I changed the output to case 1 and added unit tests. I also added the dimension prameter as suggested by @tomasstolker which allows to either stack along the wavlength or time dimension with 4D data.

tomasstolker:
Thanks, @Kiefersv, looks good to me.

I made some minor style improvements and changes to the docstrings.

Before we merge the PR, could you have a look at the docstring of `image_out_tag`? I think that it might be outdated.

Kiefersv:
Yes, it was indeed outdated. I changed it to:
`Tag of the database entry that is written as output. The shape of the output data is equal to the data from image_in_tag except, if stack is not None, the stacked dimension has a size of 1.`

If you think this is formulated too complicated `Tag of the database entry that is written as output.` could also suffice.

tomasstolker (approved):
Thanks, @Kiefersv 👍

PR #430, opened 2021-12-08
Bug fixes in Contrast Curves

### LINKED ISSUE #424: Bugs in contrast curve
Hi all,

recently I revisited the Mawet et al 2014 paper and check our implementation. Unfortunately, I found some bugs in the `contrast_limit` function:

- [ ] In the calculation of the planet self-subtraction induced by PCA:
https://github.com/PynPoint/PynPoint/blob/00c4cb5a4c412f43a6888311fc16c763f2d55a9a/pynpoint/util/limits.py#L158
The attenuation is the ratio of planet flux in the residual compared with the inserted flux. The planet flux in the residual is directly measured on the residual image:
https://github.com/PynPoint/PynPoint/blob/00c4cb5a4c412f43a6888311fc16c763f2d55a9a/pynpoint/util/limits.py#L151
The value of the residual before the planet injection is not considered. The estimate of `flux_out` should be `flux_out = flux_planet_residual - flux_residual`.

- [ ] the noise `t_noise` is estimated by using the function `false_alarm`. 
https://github.com/PynPoint/PynPoint/blob/00c4cb5a4c412f43a6888311fc16c763f2d55a9a/pynpoint/util/limits.py#L116
this function calculates the noise only based on n-1 residual elements (as for the S/R calculation). In the case of the contrast curve, we have one additional residual element as the planet is not present.

- [ ] The final contrast is calculated in Line:
https://github.com/PynPoint/PynPoint/blob/00c4cb5a4c412f43a6888311fc16c763f2d55a9a/pynpoint/util/limits.py#L161
The term `sigma*t_noise` can be interpreted as the brightness a planet needs to have in order to be detectable under the given confidence level. As explained in the Mawet paper the average of the noise residual elements should be added to this term (x in the paper sec. 3.4). Usually, this is equal to zero; but if not we get errors. @tomasstolker You once had problems estimating the contrast for residuals with a lot of negative values. I think this might be the reason. 

Cheers,

Markus

### LINKED ISSUE #431: Many dependencies are outdated
Since the last release of pynpoint many dependencies have been updated to new versions. The package is no longer compatible with the current standard version of e.g. numpy and scipy. This causes installation problems if pynpoint is used inside or along with other packages.

We should update the requirement and fix the tests.

markusbonse:
This is pull request fixes several bugs in the contrast curves modules.

The biggest change is related to this line:
https://github.com/majobodo/PynPoint/blob/ad9c415529668fdc5949f0040ef8456018a23037/pynpoint/util/limits.py#L172

the average of the noise apertures needs to be added if we want to compute the contrast curve. This is also stated in the original paper by Mawet et al 2014. In most cases this does not matter as the average is close to 0. However it can have quite some impact if we move to high PCA components.

For the calculation of the detection confidence we subtract it:
https://github.com/PynPoint/PynPoint/blob/45b3d2d9e42282d835477d8a3621833ce1769b60/pynpoint/util/analysis.py#L124

Thus, we have to add it for the contrast curves.

All tests finish without error on my local machine, but they crash on the CI.

Here are some plots of CC before and after the bug fix. 
BEFORE:
Before_bug_fix

AFTER:
After_bug_fix


majobodo:
It seems to run with python 3.7 but not with 3.8 and 3.9. Maybe this is due to the outdated dependencies in Pynpoint?

majobodo:
Tests and build works. The code coverage raises a Segfault for unknown reasons if not run with option  --collect-only. Maybe it no longer works

tomasstolker (approved):

